PR #3, opened 2021-09-17
Remove broken chart. Restore local backend chart [BW-816 pre-work]

cjllanwarne:


CromwellPostgresPapiMinikube/README.md
```diff
@@ -5,12 +5,12 @@ To run this in minikube:
 3. `minikube addons enable metallb `
```
cahrens, line 5: Do we want a version of this README in CromwellPostgresLocalMinikube as well?

PR #8, opened 2021-10-14
Ingress for Local HSQLDB chart for deployment via Leo

cjllanwarne:


charts/index.yaml
```diff
@@ -3,12 +3,32 @@ entries:
...
   - apiVersion: v2
     appVersion: 68-8e12ab5
-    created: "2021-09-21T18:48:15.630787-04:00"
+    created: "2021-10-01T17:29:10.346454-04:00"
+    description: Simple Cromwell deployment with in-memory database running jobs locally
+    digest: b8aa2050dc90a8ea6c407d6fa2e5f6ab06b28cf60d30936678d054cb15868236
+    name: CromwellHsqldbLocalMinikube
+    type: application
+    urls:
+    - https://broadinstitute.github.io/cromwhelm/charts/CromwellHsqldbLocalMinikube-0.1.2.tgz
```
rtitle, line 12: Maybe in the future we could look at using GCS as a helm repo instead of GitHub. We've seen performance issues in the past in Leo with repeatedly downloading charts from GitHub. Probably not an issue short term though..

PR #11, opened 2021-10-29
Run with Postgres database (BW-863)

cahrens:
Cromwell running with Postgres as deployed from Leonardo:

image

image

Demonstration that "replace-me" is coming through as the db password from Leo, when I (temporarily) displayed the password as an annotation.

image


cromwell-helm/templates/_helpers.tpl
```diff
@@ -49,3 +49,17 @@ Selector labels
...
+
+{{/*
+Return postgres database user password.
+Lookup the existing secret values if they exist, or generate a random value
+*/}}
+{{- define "dbPassword" -}}
+{{- if .Values.db.password }}
+    {{- .Values.db.password -}}
+{{- else }}
+    {{- $randomValue := (randAlphaNum 32) -}}
```
cahrens, line 61: Stole this from Galaxy helm charts.

cromwell-helm/templates/postgres.yaml
```diff
@@ -0,0 +1,52 @@
...
+###########################
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "app.fullname" . }}-postgres-depl
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      component: postgres
```
cahrens, line 12: I don't know if I should make these labels more complicated (like we do for Cromwell, following Galaxy examples). TBH, I don't completely understand the point of having multiple labels.

kpierre13: What are these labels used for? Do they show up somewhere or are they just for identification purposes?

cahrens: TBH, I'm still pretty confused about how labels are used. But here is the documentation-- https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/

kpierre13 (approved):

salonishah11 (approved):
LGTM 👍

PR #12, opened 2021-11-02
Add SWAGGER_BASE_PATH  env variable  [BW-862]

salonishah11:
Closes https://broadworkbench.atlassian.net/browse/BW-862


cromwell-helm/templates/cromwell.yaml
```diff
@@ -23,7 +23,7 @@ spec:
             claimName: {{ include "app.fullname" . }}-pvc
       containers:
         - name: {{ .Chart.Name }}-container
-          image: broadinstitute/cromwell:68-8e12ab5
+          image: broadinstitute/cromwell:69-ab51713-SNAP # TODO: Update this to official version once Cromwell 71 is released
```
salonishah11, line 26: Not sure why, but for some reason when Katrina built and pushed docker image from her local branch, the hash produced was this instead of `71-<hash>-SNAP`

charts/index.yaml
```diff
@@ -45,8 +45,18 @@ entries:
   cromwell:
   - apiVersion: v2
     appVersion: 68-8e12ab5
-    created: "2021-10-29T16:37:08.32811-04:00"
-    description: Simple Cromwell deployment with Postgres database running jobs locally
+    created: "2021-11-02T16:17:45.942353-04:00"
+    description: Simple Cromwell deployment with in-memory database running jobs locally
```
cahrens, line 49: This is now Postgres database instead of in-memory.

charts/index.yaml
```diff
@@ -45,8 +45,18 @@ entries:
...
+    digest: fa9c1138027d106b876900f8ef3634e1c4065910670e6116c2f4df6da46916cc
+    name: cromwell
+    type: application
+    urls:
+    - https://broadinstitute.github.io/cromwhelm/charts/cromwell-0.1.6.tgz
+    version: 0.1.6
+  - apiVersion: v2
+    appVersion: 68-8e12ab5
+    created: "2021-11-02T16:17:45.918357-04:00"
+    description: Simple Cromwell deployment with in-memory database running jobs locally
```
cahrens, line 59: Here too.

charts/index.yaml
salonishah11: Hmmm maybe I missed a step. This file was updated by running `helm repo index --url https://broadinstitute.github.io/cromwhelm/charts/ charts`. Do I need to change something else?

cahrens: Oh, I bet it's picking up that description from somewhere else then. I edited it here, but that probably was wrong. Don't worry about it for now… it's not a big deal.

cahrens: I see… it should be updated in Chart.yaml (and that's where the Cromwell version is also).

salonishah11: Thank you! I will update in Chart.yaml and regenerate the files

cahrens (approved):

charts/index.yaml
```diff
@@ -65,34 +75,23 @@ entries:
...
-  - apiVersion: v2
-    appVersion: 1.16.0
-    created: "2021-10-29T16:37:08.329601-04:00"
-    description: Simple Cromwell deployment with Postgres database
-    digest: ad93460102e6d790e3255981b1dc151b36ef578064eaff7950186cd6258f8a37
-    name: cromwellwithpostgress
-    type: application
-    urls:
-    - https://broadinstitute.github.io/cromwhelm/charts/cromwellwithpostgress-0.1.0.tgz
-    version: 0.1.0
```
kpierre13, line 86: I was wondering why this block was removed? Do we not need it to run postgres?

kpierre13 (approved):

charts/index.yaml
cahrens: No, this was something I accidentally introduced because I was helping Adam to debug something. There's no checked in cromwellwithpostgress file, so this section shouldn't actually exist.

cahrens: And to be clear, the actual helm chart we are using (which now is using Postgres) is called simply "cromwell".

PR #13, opened 2021-11-05
BW-902 - Update Cromwell version in cromwhelm chart

kpierre13:


CromwellPostgresLocalMinikube/templates/cromwell.yaml
```diff
@@ -17,7 +17,7 @@ spec:
     spec:
       containers:
         - name: cromwell-container
-          image: broadinstitute/cromwell:66-5f8a277
+          image: broadinstitute/cromwell:71-c778c33
```
cahrens, line 20: Whether or not you change this one doesn't really matter…this isn't our active helm chart.

kpierre13: Should I revert this change for the sake of consistency?

cahrens: I guess it might as well be updated.

cahrens (approved):

charts/index.yaml
```diff
@@ -45,17 +45,17 @@ entries:
   cromwell:
   - apiVersion: v2
     appVersion: 68-8e12ab5
-    created: "2021-11-02T17:06:15.214495-04:00"
+    created: "2021-11-05T15:44:19.592712-04:00"
     description: Simple Cromwell deployment with Postgres database running jobs locally
-    digest: ae59ae9fa9765891d8e15f1e477901a55cbd1346621d7b7029298323c5f5fbf5
+    digest: c64a31d9c0720cdab89ba2e61f691fbea058994bfded94e1d5fa0387a5388be1
```
cjllanwarne, line 50: Apologies if this was already discussed and decided but It feels not quite right to me that we're "changing history" by updating v0.1.6 rather than publishing a new 0.1.7 version with the new cromwell jar.

charts/index.yaml
```diff
@@ -45,17 +45,17 @@ entries:
   cromwell:
   - apiVersion: v2
     appVersion: 68-8e12ab5
```
cjllanwarne, line 47: The `appVersion` now appears to be out of date

charts/index.yaml
kpierre13: @cahrens, I also had questions about why we are "overwriting" v.0.1.6 rather than creating a v.0.1.7. Is there a specific reason?

cahrens: Because @salonishah11 temporarily checked in 0.1.6 with the snapshot Cromwell version, and we planned to just adjust it to the actual Cromwell release once that was done. If you create 0.1.7 you will also need to update Leo, and that seems unnecessarily complex to me.

kpierre13: Ah, I forgot about the SNAP version. This makes sense to me now. Thank you.

charts/index.yaml
kpierre13: I changed this in multiple places with the latest commit. Did I miss any instances?

cahrens: I don't think you should be updating the index.yaml file directly (I've made that mistake myself twice!). And I don't think you want to change the Cromwell version ID for older helm file versions… those were legitimately made with a different release of Cromwell.

cjllanwarne:
@kpierre13 I think we can close this PR now that #13 has merged?

PR #15, opened 2021-11-17
Add config file for cromwell-helm [BW-916]

cjllanwarne:


kpierre13 (approved):

charts/index.yaml
```diff
@@ -45,7 +45,17 @@ entries:
   cromwell:
   - apiVersion: v2
     appVersion: 71-c778c33
-    created: "2021-11-09T10:44:48.499635-05:00"
+    created: "2021-11-17T18:25:17.783266-05:00"
+    description: Simple Cromwell deployment with Postgres database running jobs locally
```
cahrens, line 49: Nit: might be worth changing the description to include PAPI config.

cromwell-helm/templates/cromwell-config.yaml
```diff
@@ -0,0 +1,85 @@
...
+    backend {
+      default = "Local"
+      providers {
+        PAPIv2 {
+          actor-factory = "cromwell.backend.google.pipelines.v2beta.PipelinesApiLifecycleActorFactory"
+          config {
+            // Google project
+            project = "{{ .Values.cromwellConfig.gcsProject }}"
+
+            // Base bucket for workflow executions
```
cahrens, line 40: Is this the user's bucket, or something that Leo provides?

cromwell-helm/templates/cromwell-config.yaml
```diff
@@ -0,0 +1,85 @@
...
+              // Will get overridden by email address provided in workflow options, "google_compute_service_account"
+              compute-service-account = "application-default"
+
+              // Pipelines v2 only: specify the number of times localization and delocalization operations should be attempted
+              // There is no logic to determine if the error was transient or not, everything is retried upon failure
+              // Defaults to 3
+              localization-attempts = 3
+
+              // Starting from version 2 beta of the Google Life Sciences API, this is a required attribute for creating pipelines.
+              location = "us-central1"
```
cahrens, line 68: We will want to pass the location of the Workspace. Can you create a ticket for that? With all the regionality work underway, it would be good to make sure we are following the right patterns.

cromwell-helm/values.yaml
```diff
@@ -2,6 +2,12 @@ service:
   type: ClusterIP
   port: 8000
 
+cromwellConfig:
```
cahrens, line 5: I think I would simplify these as suggested below. Since you access them via the full path, "config" doesn't need to be repeated. And it seems like the general pattern is to minimize camel-cased value names.

```
config:
  dir:
  file:
  gcsBucket:
  gcsProject:

File: cromwell-helm/templates/cromwell-config.yaml (hunk 2 above)
Comment (cjllanwarne) [reply cromwell-helm/templates/cromwell-config.yaml:40]: It's the workspace bucket, which Leo will be able to know - and therefore tell us - about

Comment (cahrens) [reply cromwell-helm/templates/cromwell-config.yaml:40]: I have forgotten… what is stored in this bucket? It surely isn't the outputs of the workflows.

File: cromwell-helm/templates/cromwell-config.yaml (hunk 3 above)
Comment (cjllanwarne) [reply cromwell-helm/templates/cromwell-config.yaml:68]: Made, and moved the declaration to `values`

File: cromwell-helm/templates/cromwell-config.yaml [hunk 5]
```diff
@@ -0,0 +1,85 @@
...
+              // Will get overridden by email address provided in workflow options, "google_compute_service_account"
+              compute-service-account = "application-default"
+
+              // Pipelines v2 only: specify the number of times localization and delocalization operations should be attempted
+              // There is no logic to determine if the error was transient or not, everything is retried upon failure
+              // Defaults to 3
+              localization-attempts = 3
+
+              // Starting from version 2 beta of the Google Life Sciences API, this is a required attribute for creating pipelines.
+              location = "{{ .Values.config.gcsRegion }}"
```
cahrens, line 68: Note though that this also impacts the endpoint-url. So perhaps that should be pulled out also into values?

cromwell-helm/templates/cromwell-config.yaml
cahrens: @cjllanwarne ^

cahrens (approved):

cromwell-helm/templates/cromwell-config.yaml
cjllanwarne: Discussed in person - yes this is the location where the app user will store their workflow outputs.

PR #18, opened 2022-01-07
[BW-939] Cromwell helm chart to run compute as the pet account using Workflow Identity

salonishah11:
References:
- https://github.com/DataBiosphere/terra-app/blob/main/terra-app-setup-chart/values.yaml
- https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity

Closes https://broadworkbench.atlassian.net/browse/BW-939


cjllanwarne (approved):
TOL do we also want to make PAPIv2 the default backend and/or remove Local?

cromwell-helm/templates/cromwell-config.yaml
```diff
@@ -79,6 +79,12 @@ data:
                 project = "{{ .Values.config.gcsProject }}"
               }
             }
+
+            virtual-private-cloud {
+              network-label-key = "vpc-network-name"
+              subnetwork-label-key = "vpc-subnetwork-name"
+              auth = "application-default"
+            }
```
kpierre13, line 87: What do these lines of code do? Why do we need them?

kpierre13 (approved):

cromwell-helm/values.yaml
```diff
@@ -9,6 +9,10 @@ config:
   gcsProject: "SUPPLIED BY LEO"
   gcsRegion: us-central1 # TODO: SUPPLY VIA LEO (BW-924)
   papiv2EndpointUrl: https://lifesciences.googleapis.com/ # TODO: SUPPLY VIA LEO... or hard-code into Cromwell? (BW-924)
+  serviceAccount:
+    name: "SUPPLIED BY LEO"
+    annotations:
```
cahrens, line 14: How did you choose this to put in annotations?

cahrens:
> TOL do we also want to make PAPIv2 the default backend and/or remove Local?

I would vote that we definitely keep local (I don't see any downside in keeping it) and perhaps leave it as default until we are comfortable doing the steps to give PAPI permissions for arbitrary users.

cromwell-helm/templates/cromwell-config.yaml
salonishah11: Explained this in mob

kpierre13: Saloni's answering of this question:

Cromwell no longer runs jobs in default networks. Now, Terra has the option to create your own network so you're not limited by the IP. So we running jobs in this particular **private** network (to be supplied) rather than the default network.

salonishah11:
I agree with Christina. CaaA works fully with local backend. If the default is Papiv2, to run a workflow we would always have to go in and add the 2 roles in IAM.

PR #21, opened 2022-04-06
Initial pass at chart for deploying to Azure [BT-603]

jgainerdewar:
I don't think the contents of this chart need to be thoroughly reviewed, this is a starting point from which we can reshape things to look more like the GCP chart and eventually integrate the two. I have deployed this to AKS from my local machine with a few manual steps (Cromwell config file upload, database initialization).

### DIFF
--- a/coa-helm/templates/csi/blobfuse-proxy.yaml
+++ b/coa-helm/templates/csi/blobfuse-proxy.yaml
@@ -1,0 +1,115 @@
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: csi-blobfuse-proxy
+  namespace: kube-system
+spec:
+  selector:
+    matchLabels:
+      app: csi-blobfuse-proxy
+  template:
+    metadata:
+      labels:
+        app: csi-blobfuse-proxy
+    spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: type
+                    operator: NotIn
+                    values:
+                      - virtual-kubelet
+      initContainers:
+        - name: prepare-binaries
+          image: mcr.microsoft.com/k8s/csi/blob-csi:v1.8.0
+          command: ['sh', '-c', "cp /blobfuse-proxy/*.deb /tmp/"]
+          volumeMounts:
+            - mountPath: /tmp
+              name: tmp-dir
+      containers:
+        - command:
+            - nsenter
+            - '--target'
+            - '1'
+            - '--mount'
+            - '--uts'
+            - '--ipc'
+            - '--net'
+            - '--pid'
+            - '--'
+            - sh
+            - '-c'
+            - |
+              set -xe
+              INSTALL_BLOBFUSE=${INSTALL_BLOBFUSE:-true}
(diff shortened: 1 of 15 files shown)


coa-helm/templates/csi/blobfuse-proxy.yaml
```diff
@@ -0,0 +1,115 @@
...
+            - '--uts'
+            - '--ipc'
+            - '--net'
+            - '--pid'
+            - '--'
+            - sh
+            - '-c'
+            - |
+              set -xe
+              INSTALL_BLOBFUSE=${INSTALL_BLOBFUSE:-true}
```
mspector, line 47: I see that "blobfuse" is some kind of Azure file system driver/enabler/thing. Am I right to guess that this chart basically serves the purpose of setting up Cromwell's Azure file system? Is there more to it, at a high level?

coa-helm/templates/csi/csi-blob-controller.yaml
```diff
@@ -0,0 +1,133 @@
...
+  namespace: kube-system
+spec:
+  replicas: 2
+  selector:
+    matchLabels:
+      app: csi-blob-controller
+  template:
+    metadata:
+      labels:
+        app: csi-blob-controller
```
mspector, line 15: I see the word "blob", and I see that "CSI" (in the context of kubernetes) can mean "container storage interface". Is it easy to explain what this container is doing? (I assume it's related to the blobfuse container above)

coa-helm/templates/csi/rbac-csi-blob-controller.yaml
```diff
@@ -0,0 +1,108 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: csi-blob-controller-sa
```
mspector, line 5: lots of controllers! I see "RBAC" stands for "role-based access control"

coa-helm/templates/tes-deployment.yaml
```diff
@@ -0,0 +1,58 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    io.kompose.service: tes
+  name: tes
```
mspector, line 6: I couldn't make an educated guess as to what this container is doing

mspector (approved):
I just left a bunch of questions -- I'll followup later if you don't feel like you want to answer them here

coa-helm/templates/tes-deployment.yaml
jgainerdewar: I'm not surprised :) This container runs the [TES](https://github.com/ga4gh/task-execution-schemas) server, which accepts tasks from Cromwell and runs them on Azure Batch. We'll talk about this in more detail at our meeting on Friday.

coa-helm/templates/csi/csi-blob-controller.yaml
jgainerdewar: Yes, this is setting up the mount of the Blob containers so that Cromwell can access them as a local filesystem. I'm not deeply familiar with the details of how the blobfuse situation is set up. It won't be our long-term solution, so I haven't dug in too deeply - I wouldn't work too hard to understand anything in this `csi` subdirectory (something I should have told you at the beginning...).

mspector: thanks! I wont :)

PR #23, opened 2022-04-20
[BW-1209] Auto update cromwell-helm chart upon merge to 'main'

salonishah11:
I tested it on my branch. Details:

- Github action [output #15](https://github.com/broadinstitute/cromwhelm/runs/6103872389?check_suite_focus=true) that updates chart version from 0.1.10 to 0.1.11 (this is because my branch didn't have latest changes from 'main' branch where the version number is 0.2.0)
- Updated `index.yaml` and chart version can be seen in [auto-chart-version-update-0.1.11 branch](https://github.com/broadinstitute/cromwhelm/compare/auto-chart-version-update-0.1.11) created during testing

Closes https://broadworkbench.atlassian.net/browse/BW-1209

### DIFF
--- a/.github/workflows/update-and-publish-helm-chart.yml
+++ b/.github/workflows/update-and-publish-helm-chart.yml
@@ -38,9 +38,10 @@ jobs:
         env:
           BROADBOT_GITHUB_TOKEN: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
         run: |
+          git checkout -b auto-chart--version-update-$NEXT_VERSION
           git diff
           git config --global user.name "broadbot"
           git config --global user.email "broadbot@broadinstitute.org"
           git commit -am "Auto update to cromwell-helm chart version to $NEXT_VERSION"
-          git push https://broadbot:$BROADBOT_GITHUB_TOKEN@github.com/broadinstitute/cromwhelm.git ss_auto_update_chart_version
+          git push https://broadbot:$BROADBOT_GITHUB_TOKEN@github.com/broadinstitute/cromwhelm.git auto-chart--version-update-$NEXT_VERSION
 
--- a/.github/workflows/update-and-publish-helm-chart.yml
+++ b/.github/workflows/update-and-publish-helm-chart.yml
@@ -35,7 +35,6 @@ jobs:
           helm package cromwell-helm/
           mv cromwell-${NEXT_VERSION}.tgz charts/cromwell-${NEXT_VERSION}.tgz
           helm repo index --url https://broadinstitute.github.io/cromwhelm/charts/ charts
-
       - name: Push changes to GitHub
         env:
           BROADBOT_GITHUB_TOKEN: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
--- a/.github/workflows/update-and-publish-helm-chart.yml
+++ b/.github/workflows/update-and-publish-helm-chart.yml
@@ -37,11 +37,9 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets._GITHUB_TOKEN }}
         run: |
-#          git checkout main
           git checkout -b auto-chart-version-update-$NEXT_VERSION
           git diff
           git config --global user.name "github-actions[bot]"
           git config --global user.email "github-actions[bot]@users.noreply.github.com"
           git commit -am "Auto update cromwell-helm chart version to $NEXT_VERSION"
-          #git push https://broadbot:$GITHUB_TOKEN@github.com/broadinstitute/cromwhelm.git main
           git push https://broadbot:$GITHUB_TOKEN@github.com/broadinstitute/cromwhelm.git auto-chart-version-update-$NEXT_VERSION
--- a/.github/workflows/update-and-publish-helm-chart.yml
+++ b/.github/workflows/update-and-publish-helm-chart.yml
@@ -13,8 +13,6 @@ jobs:
     steps:
       - name: Check out the repository
         uses: actions/checkout@v2
-        with:
-          ref: ${{ github.ref }}
 
       - name: Find out next chart version
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/update-and-publish-helm-chart.yml
```diff
@@ -0,0 +1,44 @@
...
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@v2
+
+      - name: Find out next chart version
+        run: |
+          currentVersion=$(grep 'version:' cromwell-helm/Chart.yaml | awk '{print $NF}')
+
+          # this will increment only the last version digit. For example, it will update 0.1.10 => 0.1.11
+          nextVersion=$(echo ${currentVersion} | awk -F. -v OFS=. '{$NF += 1 ; print}')
```
salonishah11, line 20: Question for reviewers: Do we want to keep updating only the patch version digit or should we move to updating the minor version digit? So go from 0.1.10 => 0.2.0 => 0.3.0?

.github/workflows/update-and-publish-helm-chart.yml
```diff
@@ -0,0 +1,44 @@
...
+          helm package cromwell-helm/
+          mv cromwell-${NEXT_VERSION}.tgz charts/cromwell-${NEXT_VERSION}.tgz
+          helm repo index --url https://broadinstitute.github.io/cromwhelm/charts/ charts
+
+      - name: Push changes to GitHub
+        env:
+          BROADBOT_GITHUB_TOKEN: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
+        run: |
+          git checkout main
+          git diff
```
cjllanwarne, line 40: What's this diff for?

.github/workflows/update-and-publish-helm-chart.yml
```diff
@@ -0,0 +1,44 @@
+name: update-and-publish-cromwell-helm-chart
+on:
+  push:
+    branches:
+      - 'main'
```
cjllanwarne, line 5: Just to confirm... when this action pushes to main, it doesn't recursively re-trigger itself?

.github/workflows/update-and-publish-helm-chart.yml
cjllanwarne: For now, this seems fine.

TOL for the future... I wonder if there's a way actions can be based on PRs merging rather than on `main` updating? If so, we could maybe parameterize the action with "type of version" and update the trigger to do something like "patch by default, minor or major if "[minor version] or [major version]" appear in the PR title?

cjllanwarne (approved):
LGTM. I tried to re-run one of your runs manually but got an error... I don't know if some bad state got saved in the action which means the re-run doesn't work if there have been subsequent pushes 🤔

.github/workflows/update-and-publish-helm-chart.yml
cjllanwarne: Or actually, even if we stay on "changes to main" we could use the most recent commit message in the same way

.github/workflows/update-and-publish-helm-chart.yml
salonishah11: That's a good question. I think that it shouldn't recursively re-trigger itself, but I will confirm this statement

.github/workflows/update-and-publish-helm-chart.yml
salonishah11: I changed it from merging based on PRs to merges to main because the [auto update Cromwell version in helm chart Github action](https://github.com/broadinstitute/cromwell/pull/6739) makes a commit directly to the `main` branch. As a result if we keep it based on PRs, we miss out on updating chart version when Cromwell version update happens.

.github/workflows/update-and-publish-helm-chart.yml
salonishah11: This is just to print the diff in the action logs. I thought it would good to have it printed in logs.

aednichols: Yeah I did the same, it's really handy to have an easy way to check changes

.github/workflows/update-and-publish-helm-chart.yml
aednichols: I think we should update the minor version.

That's what I did in my PR before we put this PR in charge of that.

My reasoning is that a Cromwell upgrade should typically be a change that adds some features and is backwards-compatible.

aednichols (commented):
Looking good to me, will thumb once we have confirmation it doesn't recurse. For example, a statement in the GH Actions documentation backed up by an experiment.

.github/workflows/update-and-publish-helm-chart.yml
salonishah11: I am still looking into this. So far Google search hasn't provided me with what I am looking for...

salonishah11: It seems that if we use [GITHUB_TOKEN to push things to the repository](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow), it will not trigger recursive workflows runs. Maybe we can use that instead of `BROADBOT_GITHUB_TOKEN`? Or are they one and the same?

aednichols (approved):
> will thumb once we have confirmation it doesn't recurse

It looks like this is hard to be 100% sure of, so in the interest of time, I think I would be OK with a live experiment (my guess is it will be fine).

If closely supervised and stopped quickly if necessary, it seems low risk.

salonishah11:
@cjllanwarne @aednichols after talking with DevOps, using the `GITHUB_TOKEN` and GitHub bot to push the changes to `main` branch seemed the easiest option to not have the workflow recursively re-trigger itself. I have updated the script and it is ready for re-review.

- Documentation: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

Testing: I had the workflow commit to a new branch when there were changes to my branch. Github-actions was able to push the changes to the new branch [auto-chart-version-update-0.1.11](https://github.com/broadinstitute/cromwhelm/tree/auto-chart-version-update-0.1.11). Output from [workflow run #17](https://github.com/broadinstitute/cromwhelm/runs/6135665361?check_suite_focus=true).

Screen Shot 2022-04-22 at 6 19 51 PM

aednichols (approved):

cjllanwarne (approved):

.github/workflows/update-and-publish-helm-chart.yml
aednichols: On a subsequent ticket I set this up intentionally as a test and I can confirm that it DOES recursively re-trigger itself.

aednichols: Screen Shot 2022-08-03 at 7 13 26 PM

aednichols: And,
```
jobs:
  update-and-publish:
    if: ( !startsWith(github.event.head_commit.message, 'Auto update cromwell-helm chart version') )
```
successfully stops it.

PR #42, opened 2022-06-24
[BT-672] Cromwell as an App on Azure rough cut

mspector:
This is the rough cut version of Cromwell as an App on Azure. It represents the fastest (not necessarily minimal) set of changes necessary to achieve functional parity with the GCP version.

Like the GCP version, the Swagger pages are not working correctly, but we have deemed this acceptable for our purposes and for the moment.


mspector:
Closed because this PR has been made obsolete by another that is coming soon.

PR #45, opened 2022-07-05
[BT-625] CoA create cromwell config

mspector:
I didn't do a good job at organizing my commits to call out the actual changes, so I'll point them out in comments.

More information: https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/

### DIFF
--- a/coa-helm/README.md
+++ b/coa-helm/README.md
@@ -9,5 +9,5 @@ You should not attempt to deploy it without understanding these manual steps (no
 
 Notes for deploying:
 - This chart depends on the Library Chart `terra-batch-libchart`. At the top level of this repository, run `helm dependency update coa-helm` to populate `coa-helm/charts` with the packaged `terra-batch-libchart` dependency.
-- Populate `local_values.yaml` (or, preferably, make your own copy to avoid accidentally committing changes).
+- Copy the contents of `local_values.template.yaml` into a new file called `local_values.yaml`, and populate this new file with the appropriate values (ask a member of the Batch team). Do not commit these new values; `local_values.yaml` should be excluded from version control by the top-level `.gitignore` file.
 - After running `helm dependency update coa-helm`, run `helm upgrade -f local_values.yaml cromwell-azure ./coa-helm`


coa-helm/templates/cromwell.yaml
```diff
@@ -0,0 +1,168 @@
...
+            - mountPath: /cromwell-workflow-logs
+              name: cromwell-workflow-logs-claim1
+            - mountPath: /workflows
+              name: workflows-claim1
+            - mountPath: /inputs
+              name: inputs-claim1
+            - mountPath: /outputs
+              name: outputs-claim1
+            - mountPath: /configuration/cromwell
+              name: cromwell-config-volume
```
mspector, line 48: this block mounts the named configMap's `data` field to the location specified by `mountPath`

coa-helm/templates/cromwell.yaml
```diff
@@ -0,0 +1,168 @@
...
+            claimName: workflows-claim1
+        - name: inputs-claim1
+          persistentVolumeClaim:
+            claimName: inputs-claim1
+        - name: outputs-claim1
+          persistentVolumeClaim:
+            claimName: outputs-claim1
+        - name: cromwell-config-volume
+          configMap:
+            name: {{ include "app.fullname" . }}-cromwell-config
```
mspector, line 74: this block defines a volume based on the configMap with the name specified

coa-helm/templates/cromwell.yaml
```diff
@@ -0,0 +1,168 @@
...
+status: {}
+
+---
+
+apiVersion: v1
+kind: Service
```
mspector, line 80: this whole service definition was originally in `service.yaml`.

coa-helm/templates/cromwell.yaml
```diff
@@ -0,0 +1,168 @@
...
+  loadBalancer: {}
+
+---
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "app.fullname" . }}-cromwell-config
+data:
+  cromwell-application.conf: |-
```
mspector, line 102: I copy/pasted this data from the `cromwell-application.conf` file that I found in the persistent volume.

coa-helm/templates/cromwell.yaml
jgainerdewar: Huh, so when we mount a config map, Helm understands that to mean we create file in the mounted directory for each item in the map, and insert that item's value into it? Or is there something here that's telling it to do that as opposed to something else?

mspector: yes! 

though I think it's a Kubernetes feature (you don't have to be using Helm, necessarily)

coa-helm/templates/cromwell.yaml
```diff
@@ -0,0 +1,168 @@
...
+            }
+          }
+        }
+      }
+    }
+
+    database {
+      db.url = "jdbc:mysql://mysqldb/cromwell_db?useSSL=false&rewriteBatchedStatements=true&allowPublicKeyRetrieval=true"
+      db.user = "cromwell"
+      db.password = "cromwell"
```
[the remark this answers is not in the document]
mspector: Good catch!

coa-helm/templates/cromwell.yaml
mspector: more info here: https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/

coa-helm/local_values.yaml
```diff
@@ -0,0 +1,28 @@
+# This is a template.
+# Don't change this file; copy it instead, and fill in the values needed for deployment.
+# Obtain these values from someone on the development team.
```
jgainerdewar, line 3: A pattern I've seen used for this is to rename this file to something like `local_values.template.yaml` and add `local_values.yaml` to `.gitignore`. Then people can copy the template and have a local working copy that they don't have to worry about accidentally checking in.

coa-helm/local_values.yaml
```diff
@@ -0,0 +1,28 @@
...
+  relayListenerConnectionString: "UNDEFINED"
+  relayListenerEndpoint: "UNDEFINED"
+  relayListenerConnectionName: "UNDEFINED"
+  relayListenerSamUrl: "UNDEFINED"
+
+db:
+  name: cromwell_db
+  port: 3306
+  user: cromwell
+  password: cromwell
```
jgainerdewar, line 22: We should leave these undefined as well, even though they won't change. The biggest requirement is that we don't want a password checked into source control.

mspector: I wondered about this -- I certainly don't mind changing these to "UNDEFINED" or something, but doing that would not actually be different in practice. The name, user, and password would just be the string "UNDEFINED"

I'll do it anyway because I agree it's better for the human reader...

coa-helm/local_values.yaml
mspector: it's funny, I considered doing exactly this... I don't know why I changed my mind. I'll make the change.

coa-helm/local_values.yaml
mspector: Actually, maybe I misunderstood: are you suggesting that I actually leave these values completely empty (thus making it invalid YAML)? 

I wonder if that would be better

jgainerdewar: No, I just mean that we shouldn't check the actual, real database connection information into source control. It's fine to check in a fake placeholder value. You can't log into any relevant database using username and password "UNDEFINED." 😄

mspector: I was thinking that leaving them blank (leaving the YAML in an invalid state) would guard against the possibility of a person accidentally deploying the default values, mistakenly thinking that "UNDEFINED" was somehow special. But maybe I'm overthinking it.

coa-helm/README.md
```diff
@@ -9,4 +9,5 @@ You should not attempt to deploy it without understanding these manual steps (no
 
 Notes for deploying:
 - This chart depends on the Library Chart `terra-batch-libchart`. At the top level of this repository, run `helm dependency update coa-helm` to populate `coa-helm/charts` with the packaged `terra-batch-libchart` dependency.
-- After running `helm dependency update coa-helm`, run `helm upgrade -f your_local_values.yaml cromwell-azure ./coa-helm`
\ No newline at end of file
+- Populate `local_values.yaml` (or, preferably, make your own copy to avoid accidentally committing changes).
```
jgainerdewar, line 12: This line should probably be updated now that we have the template.

coa-helm/templates/cromwell.yaml
```diff
@@ -0,0 +1,168 @@
...
+    system {
+      input-read-limits {
+        lines = 1000000
+      }
+    }
+
+    engine {
+      filesystems {
+        local {
+          enabled: true
```
kraefrei, line 120: I remember from one of the meetings with Microsoft there being a discussion of local filesystem access. Does this relate back to this? Should we enable local filesystem access?

jgainerdewar: Right now this instance of Cromwell thinks it's running on a local filesystem, which is actually Blob containers mounted via [blobfuse](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-how-to-mount-container-linux). But soon... BlobFilesystem!

kraefrei (approved):

PR #46, opened 2022-07-08
BW-1284 Move default path in reverse-proxy-config and update CBAS service to latest image

salonishah11:
Closes https://broadworkbench.atlassian.net/browse/BW-1284


cjllanwarne (approved):
I suspect this wasn't actually strictly necessary with the other fix made in CBAS UI, but it doesn't hurt to have it

PR #48, opened 2022-07-12
[BT-673] cromwell-helm with libchart

mspector:
This PR does not actually change any functionality!

The configurations are now inherited from `terra-batch-libchart`. Happy to explain how this works in detail if anyone is interested.


terra-batch-libchart/templates/_cbas-ui.tpl
```diff
@@ -25,7 +25,7 @@ spec:
                 path: {{ .Values.config.cbasUI.conf_file }}
       containers:
         - name: {{ .Values.batchAnalysisUI.name }}-container
-          image: us.gcr.io/broad-dsp-gcr-public/terra-batch-analysis-ui:cjl_BW-1291
+          image: us.gcr.io/broad-dsp-gcr-public/terra-batch-analysis-ui:sshah-final-check
```
salonishah11, line 28: I think it would be better to keep UI at `cjl_BW-1291` because `sshah-final-check` has some UI changes that were later removed. The latest image for UI would be published soon in [BW-1296](https://broadworkbench.atlassian.net/browse/BW-1296)

mspector: sounds good. actually, this makes me wonder if there's a better way to keep images up to date.

for now though, I'll just change it back to `cjl_BW-1291` as suggested

salonishah11: Thanks! We are also trying to automate some of these things in our [CI/CD epic](https://broadworkbench.atlassian.net/browse/BW-1280). So hopefully soon both publishing of these images and updating them in cromwhelm will happen as soon as commits are made to their respective repo's `main` branch.

salonishah11 (approved):
LGTM!

gdlex4015:
I do not see any checks (tests, code coverage, etc) run on this PR?

PR #50, opened 2022-07-19
[BT-673] make grep search for version at the start of a line only

mspector:


.github/workflows/update-and-publish-helm-chart.yml
```diff
@@ -14,7 +14,7 @@ jobs:
 
       - name: Find out next chart version
         run: |
-          currentVersion=$(grep 'version:' cromwell-helm/Chart.yaml | awk '{print $NF}')
+          currentVersion=$(grep '^version:' cromwell-helm/Chart.yaml | awk '{print $NF}')
```
mspector, line 17: Why this is needed:

with the introduction of `terra-batch-libchart`, `cromwell-helm/Chart.yaml` now has two instances of the string `"version:"`, which causes the output of grep to contain both versions separated by a newline, making the version number invalid.

Adding the `^` anchor to the grep string causes it to select for instances of `version:` only at the very beginning of lines.

jgainerdewar (approved):

breilly2 (approved):

salonishah11 (approved):
💯

gdlex4015:
I do not see any checks (tests, code coverage, etc) run on this PR?

PR #51, opened 2022-07-28
[BT-705] template cleanup

mspector:
Simple naming change, to be more consistent across variables, images, etc.

There's more cleanup that we could do, but this is a minimal set of changes that gets us the most important cleanup improvements (IMO)


salonishah11 (approved):
This LGTM. I was wondering if there was a way to keep the values for `cbas`, `cbas-ui` and `cromwell` images in a single file in some common folder? This way we just need to update 1 file instead of 2 and the CoA and CaaA on GCP will always be in sync.

mspector:
@salonishah11 , yes, I think we can do that by just splitting up the values files, and then passing each one in with the `-f` flag when we do `helm upgrade` (e.g. `helm upgrade -f common-values.yaml -f coa-helm-values.yaml ...`)

this should be pretty easy to update. I could include it in a followup commit.

mspector:
@salonishah11 , the "easy" way that I thought would work is not...

is it reasonable to keep this in mind for future updates?

salonishah11:
@mspector yeah if it is non-trivial we can add it later as an enhancement 👍

breilly2 (approved):

gdlex4015:
Any verifications set up in this repo? (tests, etc)

PR #62, opened 2022-08-09
BW-1379: Add WDS to CoA Helm

cjllanwarne:
- Moves WDS config into per-cloud charts
- Adds mysql custom startup scripts to create all required databases


coa-helm/templates/mysqldb-deployment.yaml
```diff
@@ -42,4 +42,59 @@ spec:
...
+    io.kompose.service: mysqldb
+status:
+  loadBalancer: {}
+
+---
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "app.fullname" . }}-mysql-startup
```
mspector, line 76: I've been using the convention `<app.fullname>-<service>-config` for ConfigMap names. Can we use that here?

The `-startup` suffix still makes sense above, where you define the volume name. Giving the ConfigMap and the Volume different names makes it ever so slightly easier to parse for a person reading, IMO

terra-batch-libchart/templates/_wds.tpl
```diff
@@ -72,25 +59,3 @@ spec:
 {{- define "terra-batch-libchart.wds-service" -}}
 {{- include "terra-batch-libchart.util.merge" (append . "terra-batch-libchart.wds-service.tpl") -}}
 {{- end -}}
-
-
-{{- define "terra-batch-libchart.wds-config.tpl" -}}
```
mspector, line 77: for other ConfigMaps, I think we are keeping the common elements in terra-batch-libchart even when those elements are pretty limited (e.g. see the ConfigMap for the reverse proxy).

I might suggest doing the same here -- i.e. keeping the definition, `apiVersion`, `kind`, and `metadata` here in the libchart, and overwriting the `data` in each of `coa-helm` and `cromwell-helm`

cjllanwarne:
Superseded by postgres migration tickets

PR #68, opened 2022-08-29
[BT-739] refactor charts

mspector:


terra-batch-libchart/values.yaml
```diff
@@ -0,0 +1,51 @@
+exports:
+  common:
+    cromwell:
+      name: coa-cromwell-svc
+      image: broadinstitute/cromwell:84-3a19be6-SNAP
```
mspector, line 5: how does this get automatically updated?

aednichols (commented):
Might be a good opportunity to clean up `CromwellPostgresPapiMinikube` and `CromwellPostgresLocalMinikube`, not updated in 10 months and I don't think they're used

mspector:
@aednichols by "clean up", do you mean "remove" (and eventually replace from scratch if/when someone wants to spin up a minikube deployment of this stuff)?

it does look pretty defunct

aednichols:
I guess I mean "delete if not usable", with only a cursory evaluation of "is usable"

terra-batch-libchart/values.yaml
aednichols: Are you sure it is updated automatically? `3a19be6` is [two commits behind](https://github.com/broadinstitute/cromwell/commits/develop) the latest at `79afa29`.

We do see `79afa29` [here](https://github.com/broadinstitute/cromwhelm/pull/68/files#diff-e2f694d4842de8190150ccfc1ad19f5c04ba9d06cf0ffd3e46b0ef20e8ae917cR19-R20) where its update mechanism is the linked PR.

mspector:
Having thought about this a little bit... it might be worth the effort to make them usable. We don't have a great local testing setup right now. I'll raise this with the BT team (and see if it's worth prioritizing)

terra-batch-libchart/values.yaml
mspector: nice catch!

mspector: (I'm not sure that it's updated automatically; I must be wrong about that)

mspector:
> Having thought about this a little bit... it might be worth the effort to make them usable. We don't have a great local testing setup right now. I'll raise this with the BT team (and see if it's worth prioritizing)

decided against this for now.

mspector:
@aednichols , unless you object, I think there is some value in keeping the minikube charts around, at least for now.

I think there will soon come a time when I (or someone) wants to run these in minikube. At that point, having the old files may be useful...

aednichols (approved):
LGTM 👍

cjllanwarne (approved):
Quick skim but it all looks good to me. 👍

jgainerdewar (approved):
I did a moderately close skim and this all makes sense to me, assuming it works in practice.

PR #70, opened 2022-09-26
[BT-1386] update wds image to latest version

mspector:
Note that the WDS team does not have image versioning that lines up well with WDS client versions. It seems like the best we can do is use the tags that are available.


terra-batch-libchart/values.yaml
```diff
@@ -32,7 +32,7 @@ exports:
 
     wds:
       name: wds
-      image: gcr.io/broad-dsde-dev/workspacedata
+      image: us.gcr.io/broad-dsp-gcr-public/terra-workspace-data-service:6786552
```
cjllanwarne, line 35: It would be nice to make this semantic version if one exists?

cjllanwarne (approved):
LGTM but a semantic version would be much nicer if possible

terra-batch-libchart/values.yaml
mspector: it does not :( 

https://broadinstitute.slack.com/archives/C03T815L316/p1663958552690649

PR #76, opened 2022-10-17
[WM-1462] Simple GHA check of Helm templates

mspector:


kpierre13 (approved):

.github/workflows/check-templates.yaml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - name: Check out the repository
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.BROADBOT_TOKEN }}
+      - name: Update dependencies
+        run: |
+          helm dependency update cromwell-helm/
+          helm dependency update coa-helm/
+      - name: Check templates
```
salonishah11, line 20: Based on [this Slack message](https://broadinstitute.slack.com/archives/C1EH66VCM/p1666021859611899?thread_ts=1666021055.532679&cid=C1EH66VCM), I was wondering if it would be a good idea to add a comment here stating that this prints the db password but there is no need to worry as it is not any env password but encoded version of value "SUPPLIED BY LEO". What do you think?

mspector: I like it

PR #77, opened 2022-10-17
[TOAZ-226] Don't override classpath-level WDS config

rtitle:
I'm working on deploying the `coa` chart through Leo. Before this fix, WDS operations were failing with:
```
java.lang.IllegalArgumentException: dataSource or dataSourceClassName or jdbcUrl is required.
```

It works now!

I think the reason is that the helm chart is providing an external `/etc/conf/wds.yaml` configuration file, replacing WDS `classpath:application.properties` ([source](https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/service/src/main/resources/application.properties)). See comments inline.


terra-batch-libchart/templates/_wds.tpl
```diff
@@ -26,7 +26,7 @@ spec:
               mountPath: {{ .Values.wds.conf_dir }}/{{ .Values.wds.conf_file }}
               subPath: {{ .Values.wds.conf_file }}
           env:
-            - name: SPRING_CONFIG_LOCATION
+            - name: SPRING_CONFIG_ADDITIONAL-LOCATION
```
rtitle, line 29: I think `SPRING_CONFIG_LOCATION` replaces the default config location in the classpath, whereas `SPRING_CONFIG_ADDITIONAL-LOCATION` adds an additional location. See: https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config.files
 
It feels like the latter is more what we want, so we can keep `classpath:application.yaml` as a default and only put environment-specific configs in the helm chart.

Q: should C-BAS do similar here?

terra-batch-libchart/templates/_wds.tpl
```diff
@@ -81,14 +81,7 @@ metadata:
   name: {{ include "app.fullname" . }}-wds-config
 data:
   {{ .Values.wds.conf_file }}: |-
-    spring:
```
rtitle, line 84: I don't think we actually need any config overrides for WDS, as the only environment-specific values are controlled by EVs (which the helm chart sets): https://github.com/DataBiosphere/terra-workspace-data-service/blob/main/service/src/main/resources/application.properties

terra-batch-libchart/templates/_wds.tpl
mspector: interesting - makes sense to me

mspector (approved):

cjllanwarne (approved):

PR #87, opened 2022-10-26
TOAZ-243: Also publish Cromwell on Azure to github pages

rtitle:
It looks like this repo's actions _is_ bumping the `coa-helm` chart version, but not publishing it to GitHub pages. Hoping this does the trick.

We are also not yet auto-updating Leo with `coa-helm` version upgrades, but I'm thinking we can punt that till later as the Azure stuff is still a bit in flux.


rtitle:
@cjllanwarne @mspector does this look right? Let me know, thanks

cjllanwarne (approved):
Looks good to me

mspector (approved):
yes, looks good!

PR #88, opened 2022-10-26
[WM-1471] add automerge workflow to Cromwhelm

mspector:
this workflow should cause PRs with an "automerge" label (and _without_ a "work in progress" label) to be automatically merged into this repository.

this is really part of an effort to configure this functionality in Leonardo, but we'd like to test it here first (due to lower consequences for failure).

(note that automerged PRs should not trigger other workflows -- in other words, we should not expect to see the Leo PR workflow run as a direct consequence of an automerged PR)

although this is a test, we may find this functionality useful in our repo as well.

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -1,0 +1,28 @@
+name: automerge
+on:
+  pull_request:
+    types:
+      - labeled
+      - unlabeled
+      - synchronize
+      - opened
+      - edited
+      - ready_for_review
+      - reopened
+      - unlocked
+  pull_request_review:
+    types:
+      - submitted
+  check_suite:
+    types:
+      - completed
+  status: {}
+jobs:
+  automerge:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/automerge.yml
```diff
@@ -0,0 +1,28 @@
+name: automerge
+on:
+  pull_request:
+    types:
```
kraefrei, line 4: Is this list comprehensive? Wanted to ask about why these types are picked?

kraefrei (approved):
This looks good I think, just one clarification question.

.github/workflows/automerge.yml
mspector: I think it is comprehensive -- these were the recommended types for the standard configuration of this plugin, I think: https://github.com/broadinstitute/automerge-action#usage

but your comment is prompting me to look into this in more detail. I'm like 80% sure about this.

kpierre13 (approved):

PR #101, opened 2022-10-31
WX-744 Blob switcheroo

jgainerdewar:
This PR makes several changes to the Azure version of the Cromwell/CBAS Helm chart. 
 * Replace blobfuse file access with native access in Cromwell. 
   * Previously we used six containers mounted via blobfuse
   * Now we use a single container that apps will access natively. 
 * Remove unused components
   * Trigger service
   * Blobfuse support
 * Put Cromwell's execution directory in an app-specific location in the container
 * Update to the latest version of TES (not needed for this change, but it had been a while)

See ticket for notes on the overall functionality of the app as of this changeset.


jgainerdewar:
Waiting to merge this until after https://github.com/broadinstitute/cromwell/pull/6941 goes in, it will require a small config change here.

aednichols (approved):

PR #110, opened 2022-11-04
TOAZ-242 Add samAction parameter to listener deployment

rtitle:
Provides helm support for new `samAction` parameter added here: https://github.com/DataBiosphere/terra-azure-relay-listeners/pull/22/

Leo needs to configure the sam action differently depending on whether it's a runtime or kubernetes-app.

Tested this locally using Leo `AKSManualTest`.


mspector (approved):
LGTM

(Heads up: the "create Leo PR" and "deploy new charts" workflows are currently not being triggered upon merging into main. You'll need to run those manually for now. Let me know if there are problems with this -- I'm the one that disabled them)

rtitle:
Thanks @mspector. There's actually a [Leo PR](https://github.com/DataBiosphere/leonardo/pull/2979) that goes along with this change, so I'll update the chart version as part of that.

PR #111, opened 2022-11-07
WM-1526: Increase workflow start rate

cjllanwarne:
Includes workflow start rates pre- https://github.com/broadinstitute/firecloud-develop/pull/2730, since that change was made for distributional, rather than start-rate, reasons.

### DIFF
--- a/coa-helm/templates/cromwell.yaml
+++ b/coa-helm/templates/cromwell.yaml
@@ -58,6 +58,20 @@ data:
     }
 
     system {
+      max-concurrent-workflows = 10000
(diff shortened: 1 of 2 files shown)


mspector (approved):
LGTM

FWIW: I've looked into ways to make it possible to define configs like this in parts (so that we don't need to repeat it in both charts), but the ways I have discovered seem more convoluted than they are worth in saved copy/pasting. If this starts to become a bigger problem, maybe we should revisit them.

salonishah11 (approved):

PR #121, opened 2022-11-14
[WM-1467] add test workflow

mspector:


mspector:
this PR is in service of testing the github actions needed for WM-1467, but the changes will not be kept, so I'm admin-merging it.

PR #123, opened 2022-11-14
[WM-1467] make Leo PR in update-and-publish action

mspector:
also checks for Jira ID before proceeding


.github/workflows/make_leo_pr.yml
```diff
@@ -1,122 +0,0 @@
-name: Make Leonardo PR
-
-on:
-  # workflow_run:
-  #   workflows: [update-and-publish-cromwell-helm-chart]
-  #   types: [completed]
-  workflow_dispatch:
```
cjllanwarne, line 7: TOL I wonder whether leaving this action in place, but maybe with a manual trigger and no expectation of automerging on the other side, might be a good idea? It may be nice to be able to force creation of a update PR? OTOH it might be become yet another unused thing to maintain? And we can already recreate it if we need it... so ok, I've changed my mind, let's delete this for now...

.github/workflows/update-and-publish-helm-chart.yml
```diff
@@ -1,16 +1,21 @@
...
+    branches:
+      - 'main'
 
 jobs:
   update-and-publish:
     if: ( !startsWith(github.event.head_commit.message, 'Auto update cromwell-helm chart version') )
     name: Update and publish new cromwell-helm chart
     runs-on: ubuntu-latest
     steps:
+      - name: Ensure that a Jira ID is present in the commit message
```
cjllanwarne, line 13: TOL:

- Can we have a github action that checks that PR titles have jira tags before allowing us to commit?
- Can we validate/change our repo settings to make sure the PR title becomes (at least part of) the commit message?

.github/workflows/update-and-publish-helm-chart.yml
```diff
@@ -1,16 +1,21 @@
...
+        run: |
+          JIRA_ID=$(echo '${{ github.event.head_commit.message }}' | grep -Eo '\[?[A-Z][A-Z]+-[0-9]+\]?')
```
cjllanwarne, line 15: Some thoughts:
- Does this work on multi-line messages?
- Does this work if the jira ID is embedded in `[]`s? Or in other ways not at the start of the message?
  - If we eventually add a "Does the title contain a jira ID" check, we should duplicate the same pattern matching we use here to pick it out of the title.
- Do we want an allow-list of JIRA project codes we might use (to reduce the chance of falsely identifying a random part of the message string as a jira ID)?
- This is a great candidate for "if it fails, alert the team in slack"

.github/workflows/update-and-publish-helm-chart.yml
```diff
@@ -57,3 +62,109 @@ jobs:
           git config --global user.email "broadbot@broadinstitute.org"
           git commit -am "Auto update cromwell-helm chart version to $NEXT_VERSION_GCP and coa-helm chart version to $NEXT_VERSION_COA"
           git push https://broadbot:$BROADBOT_TOKEN@github.com/broadinstitute/cromwhelm.git main
+      
+      - name: Find out latest chart version from cromwhelm repository
```
cjllanwarne, line 66: Is this step different from `- name: Find out next chart version`? Should it become more like a "check the expected new chart has appeared in the charts directory" kind of an action?

.github/workflows/update-and-publish-helm-chart.yml
```diff
@@ -57,3 +62,109 @@ jobs:
...
+          echo ::set-output name=NEW_CROMWELL_HELM_V::${NEW_CROMWELL_HELM_V}
+      
+      - name: Create leonardo PR
+        id: create-leo-pr
+        uses: actions/github-script@v6
+        with:
+          github-token: ${{ secrets.BROADBOT_TOKEN }}
+          script: |
+            const pull_response = await github.rest.pulls.create({
+              title: '$JIRA_ID Update Cromwell version to ${{ steps.update-and-push.outputs.NEW_CROMWELL_HELM_V }}',
```
cjllanwarne, line 151: Curly braces are "normally a good idea" to delimit the variable name, the colon is "what I tend to do in my own PRs". Take or leave both at your discretion:
```suggestion
              title: '${JIRA_ID}: Update Cromwell version to ${{ steps.update-and-push.outputs.NEW_CROMWELL_HELM_V }}',
```

cjllanwarne (approved):
I focused mainly on the action flow itself, rather than "did you code up the specific actions correctly", since that is presumably mostly copy/pasted from elsewhere.

Looks good to me, a couple of TOL level comments.

salonishah11 (approved):

.github/workflows/make_leo_pr.yml
mspector: I went through the same sequence of thoughts, for what it's worth. I agree that it's easy enough to recreate it...

.github/workflows/update-and-publish-helm-chart.yml
mspector: > Does this work on multi-line messages?
> Does this work if the jira ID is embedded in []s? Or in other ways not at the start of the message?

I believe: yes, yes, and yes, based on some hand tests that I did

> Do we want an allow-list of JIRA project codes we might use (to reduce the chance of falsely identifying a random part of the message string as a jira ID)?

Maybe? I'm on the fence... let's talk about this tomorrow

> This is a great candidate for "if it fails, alert the team in slack"

Agreed. I'll add it to that ticket

.github/workflows/update-and-publish-helm-chart.yml
mspector: > Should it become more like a "check the expected new chart has appeared in the charts directory" kind of an action?

It's only different in that it reads directly from the charts, rather than using the computed version number

I originally added it as part of a copy/paste from the other workflow, and saw the similarity with the earlier step, but then decided to keep it because it seemed useful for that reason alone... but I agree that it'd be even better to perform a check against the computed number.

PR #130, opened 2022-11-22
[WM-1556] set up automerge test trigger

mspector:
there are a few checks that are disabled in this PR, that will be re-enabled in subsequent PRs, and a github action that will be removed after testing is complete


mspector:
force-merging this because it's necessary for testing github actions

PR #132, opened 2022-11-22
[WM-1556] adjust automerge steps for testing

mspector:


mspector:
force merging this because it's required for testing

PR #140, opened 2022-11-22
[WM-1556] cromwhelm automerge add PR number from event payload (testing continued)

mspector:


mspector:
force-merging this for testing purposes

PR #149, opened 2022-12-07
WM-1591: Include app name in cbas-ui's CBAS URL in COA

cjllanwarne:


kpierre13 (approved):

coa-helm/templates/cbas-ui.yaml
```diff
@@ -21,6 +21,6 @@ spec:
 data:
   {{ .Values.cbasUI.conf_file }}: |-
     {
-      "cbasUrlRoot" : "{{ .Values.relay.path }}/cbas"
+      "cbasUrlRoot" : "{{ .Values.relay.path }}{{ .Values.persistence.leoAppInstanceName }}}}/cbas"
```
jgainerdewar, line 24: I think this has double closing braces?

rtitle (changes requested):
As discussed in [slack](https://broadinstitute.slack.com/archives/C02H8KALYMD/p1670507219356029?thread_ts=1670448027.893499&cid=C02H8KALYMD) I think this was a Leo oversight and would be better to fix Leo-side. Here is a ticket: https://broadworkbench.atlassian.net/browse/TOAZ-296

Suggest closing this PR

rtitle:
Leo fix: https://github.com/DataBiosphere/leonardo/pull/3065

rtitle:
The Leo side is merged so this should be fixed for new CBAS apps. Going to close this

PR #151, opened 2022-12-15
WX-859 Add config for workflow-execution-identity

kraefrei:


coa-helm/templates/cromwell.yaml
```diff
@@ -141,6 +141,7 @@ data:
             endpoint = "http://tes/v1/tasks"
             use_tes_11_preview_backend_parameters = true
             transform-blob-to-local-path = true
+	    workflow-execution-identity = "{{ .Values.config.workflowExecutionIdentity }}"
```
jgainerdewar, line 144: Haha, I think Github is yelling at you about a tab character here. 

We'll also want to update `values.yaml` and `local_values.template.yaml` in this chart.

jgainerdewar:
FYI @rtitle

jgainerdewar (approved):

rtitle (approved):
Thanks, 👍 

Leo ticket to update: https://broadworkbench.atlassian.net/browse/TOAZ-303

coa-helm/values.yaml
```diff
@@ -12,6 +12,7 @@
   dockerInDockerImageName: ""
   usePreemptibleVmsOnly: false
   drsUrl: RUNTIME_PARAMETER
+  workflowExecutionIdentity: RUNTIME_PARAMETER
```
rtitle, line 15: FWIW, I _think_ this is the same value as `identity.resourceId` (L37) which Leo is already setting -- but updating Leo to send `config.workflowExecutionIdentity` as well.

PR #152, opened 2022-12-16
AJ-676: Add workspace Id, app name to deployment labels to make finding relevant pods easier

aaronkanzer:
It is hard to determine which pod specifically correlates with a workspace when parsing through pods in an AKS cluster.

The need for this is increasing as we'd like to evaluate logs/metrics of Pods as WDS gets onboarded to Azure

### DIFF
--- a/terra-batch-libchart/templates/_labels.tpl
+++ b/terra-batch-libchart/templates/_labels.tpl
@@ -1,0 +1,5 @@
+{{- define "terra-batch-libchart-coa-helm.labels" -}}
+aadpodidbinding: {{ .Values.identity.name }}
+leoAppName: {{ .Values.persistence.leoAppInstanceName }}
+workspaceId: {{ .Values.persistence.workspaceManager.workspaceId }}
+{{- end -}}
--- a/coa-helm/templates/_labels.tpl
+++ b/coa-helm/templates/_labels.tpl
@@ -1,4 +1,4 @@
-{{- define "terra-batch-libchart-coa-helm.labels" -}}
+{{- define "coa-helm.labels" -}}
 aadpodidbinding: {{ .Values.identity.name }}
 leoAppName: {{ .Values.persistence.leoAppInstanceName }}
 workspaceId: {{ .Values.persistence.workspaceManager.workspaceId }}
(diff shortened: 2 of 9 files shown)


aaronkanzer:
@rtitle @cjllanwarne Wasn't sure where to post in Slack for `cormwhelm` PR reviews -- is there a good place to do so moving forward?

mflinn-broad (approved):
lgtm. It may also be worth considering including [these labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/). We use them for the control plane but they are also a common standard in many kubernetes native applications. There is a lot of tooling that looks for them by default. May come in handy but up to you.

aaronkanzer:
> lgtm. It may also be worth considering including [these labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/). We use them for the control plane but they are also a common standard in many kubernetes native applications. There is a lot of tooling that looks for them by default. May come in handy but up to you.

Ah good point -- I think for now I'm going to keep them as custom, as it seems this might be Leo-specific for now. Will keep tabs on this pattern though if things become more streamlined

terra-batch-libchart/templates/_labels.tpl
```diff
@@ -0,0 +1,5 @@
+{{- define "terra-batch-libchart-coa-helm.labels" -}}
+aadpodidbinding: {{ .Values.identity.name }}
```
rtitle, line 2: I don't think `aadpodidbinding` should be in terra-batch-libchart, since it's Azure-specific (and the library chart supports both clouds)

rtitle:
So I think we do set those common labels in https://github.com/broadinstitute/cromwhelm/blob/main/terra-batch-libchart/templates/_helpers.tpl

Re the specific labels in this PR, I like the idea of making a helper function, but I think the helper should live in `coa-helm` for now since they are Azure-specific labels. (Technically workspaceId and leoAppId aren't Azure-specific, but are only passed to the Azure chart for now.)

rtitle (approved):
👍  thanks!

terra-batch-libchart/templates/_labels.tpl
aaronkanzer: Updated!

PR #153, opened 2022-12-22
WX-864 Add final workflow log directory config for Cromwell

kraefrei:
This change adds a config for Cromwell in CBAS to specify a final directory to put execution logs, particularly for blob storage. This change works with a change in CBAS to supply this value to WES but also requires an outstanding change with the WES API to ensure the workflow option is used for the submission


cjllanwarne:
Pedantic comment, but it might help a future ticket searcher: this change is for cromwell config rather than CBAS

cjllanwarne (approved):

jgainerdewar:
Can you also remove the Cromwell config that attempts to define the location for these logs?

cromwell-helm/templates/cbas-api.yaml
```diff
@@ -28,6 +28,7 @@ data:
       cromwell:
         baseUri: "http://{{ include "app.fullname" . }}-cromwell-svc:8000"
         healthUri: "http://{{ include "app.fullname" . }}-cromwell-svc:8000/engine/v1/status"
+        finalWorkflowLogDir: "https://{{ .Values.persistence.storageAccount }}.blob.core.windows.net/{{ .Values.persistence.blobContainer }}/workspace-services/cbas/{{ .Values.persistence.leoAppInstanceName }}/cromwell-workflow-logs"
```
cjllanwarne, line 31: Isn't this the GCP version of the app? If so it probably doesn't want this value

salonishah11 (approved):

cjllanwarne (changes requested):

cromwell-helm/templates/cbas-api.yaml
kraefrei: I had not realized! Thank makes sense though, I have now removed it

cjllanwarne (approved):

PR #154, opened 2023-01-03
AJ-676: Adjust indents on coa-helm labels

cjllanwarne:


aednichols (approved):
What an odd feature, hard to visually notice issues (unless you run their validator)

PR #158, opened 2023-01-04
AJ-717: Remove references of WDS from coa-helm chart

aaronkanzer:
Relates to https://broadworkbench.atlassian.net/browse/AJ-717

In order to de-couple WDS and CBAS, this PR is work to remove references

See https://github.com/broadinstitute/terra-helmfile/pull/3834 as well

[AJ-717]: https://broadworkbench.atlassian.net/browse/AJ-717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/update-and-publish-helm-chart.yml
```diff
@@ -85,6 +94,11 @@ jobs:
             echo "New version ${NEW_CROMWELL_HELM_V} does not match GCP version ${NEXT_VERSION_GCP}"
             exit 1
           fi
+          
+          if [ "${NEW_CROMWELL_HELM_V}" != "${NEXT_VERSION_WDS}" ]; then
+            echo "New version ${NEW_CROMWELL_HELM_V} does not match WDS version ${NEXT_VERSION_WDS}"
```
calypsomatic, line 99: I know this is just the same as the line above, but what's the purpose here?  To keep all the charts in the same version even though they're separate?

WDSPostgresAzureMinikube/README.md
```diff
@@ -0,0 +1,20 @@
...
+3. First (even though this is not the first step, or maybe it is if you already have `minikube` running locally), run `kubectl config use-context minikube`. Ensure that you are assigned locally to the right k8s context so that
+you do not push updates to any other k8s cluster.
+
+4. Then, make whatever changes you'd like in `wds-azure-helm`
+
+5. After making your changes, bump the semantic version in the `wds-azure-helm/Chart.yaml` to make sure your new chart is unique.
+
+6. Run `helm package wds-azure-helm/ -d charts/` -- this will create the packaged .tgz chart with other charts in the `cromwhelm` repository.
+
+7. Run `helm install charts/wds-azure-helm-<semantic-version>.tgz --generate-name` -- wait until you see a successful `DEPLOYED`
```
calypsomatic, line 16: I got this error: `Error: INSTALLATION FAILED: non-absolute URLs should be in form of repo_name/path_to_chart, got: generate-name`, is this instruction correct?

terra-batch-libchart/templates/_reverse-proxy.tpl
```diff
@@ -115,12 +115,10 @@ data:
         }
         {{ end }}
 
-        {{- if .Values.wds.enabled }}
         location /wds/ {
```
calypsomatic, line 118: We don't need to check if wds is enabled because it's always enabled/deployed, or why?

WDSPostgresAzureMinikube/README.md
calypsomatic: I had a space after the `--`!  My bad.

.github/workflows/update-and-publish-helm-chart.yml
aaronkanzer: It seems we move in lockstep in regards to semantic versioning (see other examples under `charts/`)

In terms of why we do this....I'm not sure....haha

aaronkanzer:
> Higher-level comments:
> 
> * after thinking about it a bit, I feel like the WDS chart should be broken out into its own repo. I don't see a reason to keep WDS and CBAS/Cromwell charts tied together and version-bumped by the same actions. I realize this makes for a nontrivial amount of more work.
> * if the only reason WDS needs the reverse proxy is to support relay-listener connections, I'd love to remove the reverse proxy from the WDS app and update Leo/listener to do the right thing. That'll be one less container to deploy into a user's cluster.

Sounds good on both! I'll break this down into a separate repo

As for the Leo work, been WIP on that today -- hopefully should have more to update on that in the near future

jordansafer (approved):

PR #167, opened 2023-01-17
WX-861 Increase IO staleness to match expected duration

aednichols:
Our default staleness timeout is [6 seconds](https://github.com/broadinstitute/cromwell/blob/develop/core/src/main/resources/reference.conf#L171-L190) and we are exceeding it by ~5 seconds as indicated by `I/O command 4.584 seconds stale...`

This PR adds those together and add a ~2x variability margin. The goal is to prevent slow but normal IO from hitting an excessively sensitive backpressure trigger.


jgainerdewar (commented):
Do you anticipate any possible weirdness from having this set too high? What symptoms would we expect to see if it were WAY too high?

aednichols:
I believe the issue that originally prompted backpressure was exhaustion of threads and thus inability to schedule any new work on the CPU. That was on multi-tenant Cromwell with 10s of thousands of IO commands.

My intuition is that in order to have problems we'd need a very heavily loaded Cromwell and increase the timeout from 6 seconds to 1000 seconds or infinity. 6 seconds was chosen based on the 5-second time we set for GCS IO commands, which is obviously no longer the filesystem of interest.

jgainerdewar (approved):
Thanks for the explanation!

PR #178, opened 2023-01-26
AJ-824: Provide coa-enabled flag to easily enable and disable k8s api-resources

aaronkanzer:
Relates to https://broadworkbench.atlassian.net/browse/AJ-824

In an effort to be able to easily configure what k8s `api-resources` are created upon applying the `coa-helm` Helm chart, this PR includes simple logic around resource definitions.


coa-helm/templates/cbas-api.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.cbas.coa-enabled }}
```
rtitle, line 1: Should we just make the `{{- if }}` block surround the whole file? Seems like we'd always want to control the deployment+service+configmap as a unit.

rtitle: Eh, don't feel that strongly

rtitle (commented):
LGTM, helm linting is sad though

terra-batch-libchart/values.yaml
```diff
@@ -4,6 +4,7 @@ exports:
       name: coa-cromwell-svc
       image: broadinstitute/cromwell:85-46a7918
       port: 8000
+      coaEnabled: false
```
jordansafer, line 7: If think we should start with these enabled by default, and allow Leo to override and disable these apps. This way, we don't impact CBAS & cromwell apps testing in dev. This won't address the WDS challenges testing in dev

jordansafer (approved):
Looks good, see my concern about defaulting the other apps to false.

terra-batch-libchart/values.yaml
aaronkanzer: Are you saying like referencing a `RUNTIME_PARAMETER` instead?

coa-helm/templates/cbas-api.yaml
aaronkanzer: Yeah the `---` essentially declares a new k8s resource, even though they are in the same yaml file -- e.g. there are 3 different k8s resources in that file

rtitle:
Going to quickly test this from Leo…

rtitle:
@aaronkanzer I deployed this PR and unfortunately hit some issues. WDS never actually deployed:
```
kubectl get pod -n dxqoyd-coa-ns                                                     
NAME                                                       READY   STATUS             RESTARTS      AGE
coa-dxqoyd-coa-rls-postgres-depl-96b4c99dd-fl2zv           1/1     Running            0             3m13s
coa-dxqoyd-coa-rls-reverse-proxy-deploy-5f958fff67-hsprp   0/1     CrashLoopBackOff   4 (58s ago)   3m13s
setup-dxqoyd-coa-rls-relaylistener-depl-76fd5cf875-v926t   1/1     Running            0             3m17s
```

And nginx-reverse-proxy is complaining abut no cromwell service (which makes sense):
```
kubectl logs -n dxqoyd-coa-ns   coa-dxqoyd-coa-rls-reverse-proxy-deploy-5f958fff67-hsprp
2023/01/26 21:28:57 [emerg] 1#1: host not found in upstream "coa-dxqoyd-coa-rls-cromwell-svc" in /etc/nginx/nginx.conf:13
nginx: [emerg] host not found in upstream "coa-dxqoyd-coa-rls-cromwell-svc" in /etc/nginx/nginx.conf:13
```

terra-batch-libchart/values.yaml
rtitle: Well, Leo can always override whatever is in `values.yaml`. Ideally though I think we'd like the dev config to reflect prod so we can test it, I think. Maybe we could have WDS-only in dev, with the ability to add other services via config change in BEEs?

rtitle:
Ah, also realizing that Leo status-checks CBAS, and expects it to be up. 🤔

coa-helm/templates/cbas-api.yaml
```diff
@@ -1,24 +1,29 @@
-{{ include "terra-batch-libchart.cbas-api-deploy" (list . "coa-helm.cbas-api-deploy") }}
 {{ define "coa-helm.cbas-api-deploy" }}
+{{- if .Values.cbas.coaEnabled }}
```
rtitle, line 2: I think this pattern might work instead:
```
{{- if .Values.cbas.coaEnabled }}
{{ include "terra-batch-libchart.cbas-api-deploy" (list . "coa-helm.cbas-api-deploy") }}
{{ end }}
{{ define "coa-helm.cbas-api-deploy" }}
# Overrides go here
spec:
  template:
    metadata:
      labels:
    {{ include "coa-helm.labels" $ | nindent 8 }}
{{ end }}
```

This way the definition (function) is still there -- but is gated from executing via the `if` block.

terra-batch-libchart/values.yaml
rtitle: Needs a bit more testing, but opened this Leo PR to control which coa services are enabled: https://github.com/DataBiosphere/leonardo/pull/3128

Leo also needs to know which services to status-check, so it makes sense for Leo to control it, I think. This way we could also enable the other ones with a Leo config change, instead of a new chart version.

rtitle: OK after discussion, I think we'll keep enabling all the services in dev, and use a BEE to test the WDS-only configuraiton. Will update Leo PR.

rtitle (changes requested):
Just requesting changes to change the conditional logic per [this comment](https://github.com/broadinstitute/cromwhelm/pull/178/files#r1088526982), and also add conditional checks to the nginx-reverse-proxy conf.

coa-helm/templates/cbas-api.yaml
aaronkanzer: Unfortunately, I don't think the syntax of Helm templates allows us to have an `if` block at the top, rather, it must be a `define`

See here for more details: https://broadinstitute.slack.com/archives/C02GJ782BAA/p1674758933800709?thread_ts=1674756048.634379&cid=C02GJ782BAA

aaronkanzer: (Here is the direct Github explanation as well: https://github.com/helm/helm/issues/10395#issuecomment-978104760)

aaronkanzer:
> @aaronkanzer I deployed this PR through Leo (no Leo changes except the chart version) and unfortunately hit some issues. WDS never actually deployed:
> 
> ```
> kubectl get pod -n dxqoyd-coa-ns                                                     
> NAME                                                       READY   STATUS             RESTARTS      AGE
> coa-dxqoyd-coa-rls-postgres-depl-96b4c99dd-fl2zv           1/1     Running            0             3m13s
> coa-dxqoyd-coa-rls-reverse-proxy-deploy-5f958fff67-hsprp   0/1     CrashLoopBackOff   4 (58s ago)   3m13s
> setup-dxqoyd-coa-rls-relaylistener-depl-76fd5cf875-v926t   1/1     Running            0             3m17s
> ```
> 
> And nginx-reverse-proxy is complaining abut no cromwell service (which makes sense):
> 
> ```
> kubectl logs -n dxqoyd-coa-ns   coa-dxqoyd-coa-rls-reverse-proxy-deploy-5f958fff67-hsprp
> 2023/01/26 21:28:57 [emerg] 1#1: host not found in upstream "coa-dxqoyd-coa-rls-cromwell-svc" in /etc/nginx/nginx.conf:13
> nginx: [emerg] host not found in upstream "coa-dxqoyd-coa-rls-cromwell-svc" in /etc/nginx/nginx.conf:13
> ```

Is something in `Leo` blocking `WDS` from deploying here? I'm slightly stumped why the Helm chart wouldn't bring the WDS `Deployment` up to life...

coa-helm/templates/cbas-api.yaml
rtitle: The above works for me in `helm template`.

I think that issue is different -- here the `define` isn't _contained_ in the if, they are independent. (You could do define..end then if..end if you want).

The define doesn't do anything by itself -- the `include` is what actually adds the template (and merges it with the base chart).

rtitle:
> I'm slightly stumped why the Helm chart wouldn't bring the WDS Deployment up to life...

I think because of the above bug in the templates. Before this PR it was:
```
{{ include the-deployment }}
{{ define the-deployment }} 
some stuff
{{ end }}
```

In this PR:
```
{{ define the-deployment }}
{{ if enabled }}
{{ include the-deployment }}
{{ end }}
some stuff
{{ end }}
```
.. which does nothing, because the top-level function is never called. We should change it to:
```
{{ if enabled }}
{{ include the-deployment }}
{{ end }}
{{ define the-deployment }}
some stuff
{{ end }}
```

aaronkanzer:
> 

@rtitle 

Ah I see 🎉 updated the PR here, hope it works! 🤞 

Thanks for suggesting the fix here.

coa-helm/templates/cbas-api.yaml
aaronkanzer: Updated!

coa-helm/templates/reverse-proxy.yaml
```diff
@@ -1,4 +1,6 @@
-{{ include "terra-batch-libchart.reverse-proxy-deploy" (list . "coa-helm.reverse-proxy-deploy") }}
+{{- if .Values.cromwell.coaEnabled }}
```
rtitle, line 1: I don't think we want this here -- probably conditionally enable each proxy path in the conf file itself (in the base chart).

aaronkanzer: Ah right -- updated the nginx conf in `terra-batch-libchart/templates/_reverse-proxy.tpl` -- I _think_ I included everything in the conditional related to `cromwell`, but I think you would know much better than I haha -- let me know if anything else should be moved out of the conditional

coa-helm/templates/cbas-api.yaml
```diff
@@ -1,24 +1,29 @@
...
 {{ include "terra-batch-libchart.cbas-api-deploy" (list . "coa-helm.cbas-api-deploy") }}
+{{ end }}
 {{ define "coa-helm.cbas-api-deploy" }}
 # Overrides go here
 spec:
-  template:
-    metadata:
-      labels:
-    {{ include "coa-helm.labels" $ | nindent 8 }}
+template:
```
rtitle, line 7: Spacing looks wrong here?

coa-helm/templates/cbas-api.yaml
```diff
@@ -1,24 +1,29 @@
...
+  metadata:
+    labels:
+  {{ include "coa-helm.labels" $ | nindent 8 }}
 {{ end }}
 
 ---
 
-{{ include "terra-batch-libchart.cbas-api-service" (list . "coa-helm.cbas-api-service") }}
 {{ define "coa-helm.cbas-api-service" }}
+{{- if .Values.cbas.coaEnabled }}
```
rtitle, line 16: This (and others) still need to be fixed w.r.t. the above pattern

coa-helm/templates/reverse-proxy.yaml
```diff
@@ -1,4 +1,4 @@
-{{ include "terra-batch-libchart.reverse-proxy-deploy" (list . "coa-helm.reverse-proxy-deploy") }}
+{{ include "terra-batch-libchart.cromwell-deploy" (list . "coa-helm.cromwell-deploy") }}
```
rtitle, line 1: Change back to `reverse-proxy-deploy`

terra-batch-libchart/templates/_reverse-proxy.tpl
```diff
@@ -82,34 +82,34 @@ data:
...
           alias /www/data/index.html;
         }
 
         # Proxying to other hosts by subpath:
-        location /cbas/ {
-          proxy_pass http://{{ include "app.fullname" . }}-cbas-svc:8080/;
-        }
-        location /cromwell/ {
-          proxy_pass http://{{ include "app.fullname" . }}-cromwell-svc:8000/;
+        if ({ .Values.cromwell.coaEnabled } = 'true') {
```
rtitle, line 91: This gates everythong on `.Values.cromwell.coaEnabled` -- think we want to change cbas/wds/cromwell for each `location`?

terra-batch-libchart/values.yaml
rtitle: Maybe default these all to true, given that Leo will be environmentally controlling it

terra-batch-libchart/templates/_reverse-proxy.tpl
aaronkanzer: I updated the logic to surround each `location` -- let me know if that's what you had in mind!

coa-helm/templates/reverse-proxy.yaml
aaronkanzer: Reverted!

coa-helm/templates/cbas-api.yaml
aaronkanzer: Updated!

coa-helm/templates/cbas-api.yaml
aaronkanzer: Hmmm I think my IDE did something funky haha -- should be fixed!

coa-helm/templates/cromwell.yaml
```diff
@@ -36,16 +38,18 @@ status: {}
 {{ end }}
 
 ---
-
-{{ include "terra-batch-libchart.cromwell-service" (list . "coa-helm.cromwell-service") }}
+{{- if .Values.cromwell.coaEnabled }}
+{{ include "terra-batch-libchart.cromwell-config" (list . "cromwell-helm.cromwell-config") }}
+{{ end }}
```
davidangb, line 43: this changed `terra-batch-libchart.cromwell-service` to `terra-batch-libchart.cromwell-config` - a cut-and-paste error maybe?

aaronkanzer: great catch here -- updated!

coa-helm/templates/cromwell.yaml
```diff
@@ -36,16 +38,18 @@ status: {}
 {{ end }}
 
 ---
-
-{{ include "terra-batch-libchart.cromwell-service" (list . "coa-helm.cromwell-service") }}
+{{- if .Values.cromwell.coaEnabled }}
+{{ include "terra-batch-libchart.cromwell-service" (list . "cromwell-helm.cromwell-service") }}
```
davidangb, line 42: I think this still isn't right, the line prior to this PR was:
```suggestion
{{ include "terra-batch-libchart.cromwell-service" (list . "coa-helm.cromwell-service") }}
```

aaronkanzer: updated!

davidangb (approved):
I am :+1: on this!

I know there is ongoing work (Rob?) to test/validate this, and to align on what the default enabled/disabled values should be in here.

aaronkanzer:
> I am 👍 on this!
> 
> I know there is ongoing work (Rob?) to test/validate this, and to align on what the default enabled/disabled values should be in here.

Yes @rtitle is testing (I know he is in the policy meetings today) -- going to try to use his BEE later to test these changes after updating the Terra UI work today

rtitle:
The conditional stuff seems to be working, but getting in `reverse-proxy` pod:
```
2023/01/31 01:47:53 [emerg] 1#1: invalid condition "(" in /etc/nginx/nginx.conf:16
nginx: [emerg] invalid condition "(" in /etc/nginx/nginx.conf:16
```
Looking, may push a commit to this PR

rtitle (approved):
I just pushed a commit to fix the nginx conf -- basically using ctmpl `{{ if }}` instead of nginx `if`. It works great now -- I used Leo to deploy all the services and WDS-only and both modes worked. 🎉

rtitle:
Leo PR to control the enabled services on the Leo side: https://github.com/DataBiosphere/leonardo/pull/3128/

firecloud-develop PR to enable all services in dev/BEE; and WDS-only otherwise: https://github.com/broadinstitute/firecloud-develop/pull/3228

They should all be good to go I think; can merge them in tomorrow after CI passing.

rtitle:
I went ahead and merged this; preparing the Leo side.

PR #180, opened 2023-01-30
AJ-822 add resource limits

calypsomatic:
See [AJ-822](https://broadworkbench.atlassian.net/browse/AJ-822)
Add initial requests and limits for CPU and memory for the CBAS, Cromwell, and WDS containers.  I welcome comments and suggestions on what are sensible numbers to use.  In most cases I've seen the numbers I have here are extremely conservative - most cpu values I've seen for containers are <~100m, memory <~256Mi - but I have seen containers that leap up to 600m+ cpu and 600Mi+ memory - unclear if we can expect those values to be necessary in normal usage or if something pathological was happening.

[AJ-822]: https://broadworkbench.atlassian.net/browse/AJ-822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/coa-helm/Chart.yaml
+++ b/coa-helm/Chart.yaml
@@ -1,6 +1,6 @@
 name: cromwell-on-azure
 description: A cromwell deployment designed to run on Azure with Postgres. 
-version: 0.2.188
+version: 0.2.187
 apiVersion: v2
 dependencies:
   - name: terra-batch-libchart
--- a/coa-helm/Chart.yaml
+++ b/coa-helm/Chart.yaml
@@ -1,6 +1,6 @@
 name: cromwell-on-azure
 description: A cromwell deployment designed to run on Azure with Postgres. 
-version: 0.2.187
+version: 0.2.188
 apiVersion: v2
 dependencies:
   - name: terra-batch-libchart
--- a/terra-batch-libchart/templates/_cbas-api.tpl
+++ b/terra-batch-libchart/templates/_cbas-api.tpl
@@ -21,6 +21,13 @@ spec:
           imagePullPolicy: Always
           ports:
             - containerPort: 8080
+          resources:
+            requests:
+              memory: "256Mi"
+              cpu: "50m"
+            limits:
+              memory: "1024Mi"
+              cpu: "200m"
           volumeMounts:
             - name: {{ include "app.fullname" . }}-cbas-config
               mountPath: {{ .Values.cbas.conf_dir }}/{{ .Values.cbas.conf_file }}
--- a/coa-helm/templates/cromwell.yaml
+++ b/coa-helm/templates/cromwell.yaml
@@ -18,7 +18,13 @@ spec:
           tty: true
           ports:
             - containerPort: {{ .Values.cromwell.port }}
-          resources: {}
+          resources:
+            requests:
+              memory: "256Mi"
+              cpu: "50m"
+            limits:
+              memory: "1024Mi"
+              cpu: "200m"
           volumeMounts:
             - mountPath: /cromwell-tmp
               name: cromwell-tmp-claim
--- a/terra-batch-libchart/templates/_wds.tpl
+++ b/terra-batch-libchart/templates/_wds.tpl
@@ -21,6 +21,13 @@ spec:
           imagePullPolicy: Always
           ports:
             - containerPort: 8080
+          resources:
+            requests:
+                memory: "768Mi"
+                cpu: "50m"
+            limits:
+                memory: "1536Mi"
+                cpu: "200m"
           volumeMounts:
             - name: {{ include "app.fullname" . }}-wds-config
               mountPath: {{ .Values.wds.conf_dir }}/{{ .Values.wds.conf_file }}
--- a/terra-batch-libchart/templates/_wds.tpl
+++ b/terra-batch-libchart/templates/_wds.tpl
@@ -23,10 +23,10 @@ spec:
             - containerPort: 8080
           resources:
             requests:
-                memory: "768Mi"
+                memory: "256Mi"
                 cpu: "50m"
             limits:
-                memory: "1536Mi"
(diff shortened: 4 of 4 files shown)


davidangb (approved):
thank you! These seem like sensible limits to start with, and we can always tweak them over time as we gather more data.

coa-helm/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: cromwell-on-azure
 description: A cromwell deployment designed to run on Azure with Postgres. 
-version: 0.2.187
+version: 0.2.188
```
aaronkanzer, line 3: I think this might auto-bump via `Broadbot` :D

terra-batch-libchart/templates/_cbas-api.tpl
```diff
@@ -21,6 +21,13 @@ spec:
           imagePullPolicy: Always
           ports:
             - containerPort: 8080
+          resources:
```
aaronkanzer, line 24: Do we know what the `cromwell-helm` Helm chart is used for?

It seems that this k8s template would also be pulled into `cromwell-helm` (it inherits from `terra-batch-libchart.cbas-api-deploy` as well.

It might be worthwhile following the same pattern as we have in `coa-helm/templates/cromwell.yaml` -- where these changes would only be self-contained to `coa-helm`

coa-helm/Chart.yaml
calypsomatic: That makes sense!  I'll undo that change

terra-batch-libchart/templates/_cbas-api.tpl
calypsomatic: No I'm afraid I don't entirely understand the directory structure of this project!  I made the cromwell changes in `coa-helm/templates/cromwell.yaml` because that was where the cromwell container specs are defined, but for cbas and wds, the yamls in `coa-helm/templates` don't define container specs but pull them in from the terra-batch-libchart templates, so that's where I made changes.

aaronkanzer: Ah seems changes to `cromwell-helm` haven't been made in a while -- wondering perhaps if `coa-helm` is a fork or something of it....

anyways, sounds good!

aaronkanzer (approved):

terra-batch-libchart/templates/_cbas-api.tpl
jgainerdewar: The naming is, uh, not the best (I can say that because it's largely my fault). `cromwell-helm` is the version of the chart for deploying the app in GCP/GKE, and was the original chart in this repo. `coa-helm` is for Azure/AKS.

jgainerdewar: If it's easy enough it feels safer to override the values and leave the GCP app as-is. It's a very different deployment--every workspace gets its own GKE cluster, no idea what the specs are--and we don't want to worry about testing things there as we iterate on the Azure resources.

calypsomatic: Sorry trying to keep up - so instead of updating templates in `terra-batch-libchart` I should only be changing files in `coa-helm`?

davidangb: This PR sets limits for `coa-helm/templates/cromwell.yaml`, and leaves `cromwell-helm/templates/cromwell.yaml`. So I think this _is_ only addressing Azure and leaves GCP/GKE untouched. Is that correct?

aaronkanzer: for `cromwell.yaml` yes, but for all other changes, specifically `cbas-api` in this PR (since the changes are being applied at the `terra-batch-libchart` level, it would affect cromwell-helm

cjllanwarne (approved):
Looks good for now but as a warning for the future, Cromwell is likely to want **significantly** more than this once workflow runs start scaling up.

My thought would be to apply this in the templates rather than directly in coa-helm where possible, but not to overthink it if that's impossible for whatever reason

jgainerdewar:
I said else where it's better not to modify the GCP deploy, but I bow to @cjllanwarne 's greater experience in that area. 🙇

terra-batch-libchart/templates/_cbas-api.tpl
jgainerdewar: We'd probably save ourselves a headache later by being consistent with this - either set everything in the library chart or set everything in the `coa-helm` chart.

jgainerdewar (approved):

terra-batch-libchart/templates/_cbas-api.tpl
rtitle: Basically `terra-batch-libchart` contains defaults, and `cromwell-helm` (GCP) and `coa-helm` (Azure) contain overrides. The yamls are merged and the overrides take precedence.

So updating `terra-batch-libchart` would apply to both clouds (unless it's overridden); but updating coa-helm would only affect Azure.

I agree it's probably good to be consistent, and probably safer at this point to only update Azure limits, since that's the one we're really testing.

rtitle (changes requested):
Requesting changes per https://github.com/broadinstitute/cromwhelm/pull/180/files#r1091371163 to update the coa-helm chart instead of the libchart (to only apply to Azure).

I think you can probably do that by putting 
```
spec:
  template:
    spec:
      containers:
        resources:
          etc
```
in the coa-helm files where it says `# Overrides go here`.

The actual memory/CPU limits look good to me for now; we can always iterate on them as time goes on.

rtitle (approved):
I just read the above comments -- if others are okay with setting it in libchart it's okay with me :)

PR #181, opened 2023-01-30
AJ-827: increase client body size to 50M for WDS

davidangb:
Default setting for `client_max_body_size` in nginx is 1M, apparently.

This restricts the size of TSV uploads for WDS. In this PR, I increase that limit to 50M.

We may want to pipe this through values at some point instead of hardcoding it?


calypsomatic (approved):
Tested it out and can now upload large(r) tsvs!

aaronkanzer (approved):

PR #184, opened 2023-02-02
[IA-3656] update deprecated k8 networking api version to v1

LizBaldo:
Relevant context can be found here: https://broadinstitute.slack.com/archives/C1EH66VCM/p1675346412119639

But the TL;DR is that the `networking.k8s.io/v1beta1` has been deprecated after k8s 1.21 and needs to be updated.


LizBaldo:
I would love to @rtitle! The instructions on the README were a bit daunting I must admit, i will find some time on your calendar, thanks!

PR #185, opened 2023-02-03
[IA-4021] fix the ingress path syntax for new v1 endpoint

LizBaldo:
Turns out that there was a few breaking changes when moving from the v1beta to the V1 endpoint:
https://github.com/kubernetes/kubernetes/pull/89778
The pathType cannot be defaulted anymore, it looks like `Prefix` would be the correct option - https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types
JIRA ticket: https://broadworkbench.atlassian.net/browse/IA-4021

### DIFF
--- a/cromwell-helm/templates/ingress.yaml
+++ b/cromwell-helm/templates/ingress.yaml
@@ -28,9 +28,12 @@ spec:
         paths:
           {{- range .paths }}
           - path: {{ . }}
+            pathType: Prefix
             backend:
-              serviceName: {{ $fullName }}-reverse-proxy-service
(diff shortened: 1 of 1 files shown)


rtitle (approved):
Wow, nice find.

Was checking Leo which dynamically sets some `ingress` fields for GKE: https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/GKEInterpreter.scala#L1454-L1463

…but it _seems_ like it'll still work.

jgainerdewar (approved):

PR #188, opened 2023-02-06
[AJ-786] increasing size limits for some large existing tables

jordansafer:
When load testing WDS, I found some current large TSVs are too large to upload. For example, a representative sample of a CDC/theiagen table. This PR increases limits by 10x, so all of these tables in the ~100 MB range can be uploaded. For larger tables (500MB+) like the CDC california table, we should explore alternate strategies for uploading such large TSVs.

Currently, a 50 MB TSV takes a bit over 30 seconds to upload, and a 500 MB takes over 5 minutes. If we see a usecase of larger uploads, we should consider UX (ie giving time estimates for long async uploads?).


terra-batch-libchart/templates/_reverse-proxy.tpl
```diff
@@ -116,7 +116,7 @@ data:
 
         {{- if .Values.wds.coaEnabled }}
         location /wds/ {
-          client_max_body_size 50M;
+          client_max_body_size 500m;
```
aaronkanzer, line 119: Assuming that the `M` is not case sensitive?

jordansafer: Ah. Apparently not- I got the lowercase m from the nginx documentation for client_max_body_size, I didn't even realize I had changed it!

aaronkanzer (approved):

calypsomatic (approved):

PR #193, opened 2023-02-10
[IA-4039] Make cbas/ui/wds/cromwell disable-able on the GCP side

rtitle:
https://broadworkbench.atlassian.net/browse/IA-4039

Separate from https://github.com/broadinstitute/cromwhelm/pull/192 to not muddy the waters.

We previously made things disablable for Azure -- doing similar for GCP.

### DIFF
--- a/coa-helm/templates/cbas-api.yaml
+++ b/coa-helm/templates/cbas-api.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.cbas.coaEnabled }}
+{{- if .Values.cbas.enabled }}
 {{ include "terra-batch-libchart.cbas-api-deploy" (list . "coa-helm.cbas-api-deploy") }}
 {{ end }}
 {{ define "coa-helm.cbas-api-deploy" }}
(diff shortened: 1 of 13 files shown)


terra-batch-libchart/values.yaml
```diff
@@ -4,7 +4,7 @@ exports:
       name: coa-cromwell-svc
       image: broadinstitute/cromwell:85-46a7918
       port: 8000
-      coaEnabled: true
+      enabled: true
```
rtitle, line 7: Changed `coaEnabled` to just `enabled` in the lib chart, and make it control both sub-charts. Default all to `true` -- Leo will enable/disable a different set of services depending on environment/cloud.

rtitle:
Tested this on a BEE (GCP and Azure) and seems to work. Requesting thumbs

LizBaldo (approved):

coa-helm/templates/cbas-api.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.cbas.coaEnabled }}
+{{- if .Values.cbas.enabled }}
```
aaronkanzer, line 1: Not sure if a corresponding Leo PR is coming up, but might want to change these values? https://github.com/DataBiosphere/leonardo/pull/3128/files#r1103089975

jgainerdewar (approved):

rtitle:
Tested this on a BEE for GCP and Azure

coa-helm/templates/cbas-api.yaml
rtitle: Yup! I did that locally - going to update the generated PR when it comes

rtitle:
Hopefully this will work because there's a JIRA ticket!

PR #200, opened 2023-02-27
[RW-8626] Enable Cromwell call caching for CaaA on GCP

yonghaoy:
AoU needs to enable call caching and it is must have for AoU launching CaaA for GCP. 
I see we already create a postgres DB in that cluster for Cromwell, seems this is the only thing left.

### DIFF
--- a/cromwell-helm/templates/cromwell.yaml
+++ b/cromwell-helm/templates/cromwell.yaml
@@ -74,6 +74,12 @@ data:
         per = 1 seconds
       }
     }
+    
+    call-caching {
+      enabled = true
+      invalidate-bad-cache-results = true
+    }
+
     google {
 
       application-name = "cromwell"


cromwell-helm/templates/cromwell.yaml
```diff
@@ -74,6 +74,12 @@ data:
         per = 1 seconds
       }
     }
+    
+    call-caching {
```
yonghaoy, line 78: not sure if this should be within `system` block...

jgainerdewar: No, it's correct to have it here as a top-level block.

jgainerdewar (approved):
This should enable call caching within a running instance of the app. As far as I know we don't currently persist data in the database across app restarts, so call caching across app "pauses" may not work.

yonghaoy:
> This should enable call caching within a running instance of the app. As far as I know we don't currently persist data in the database across app restarts, so call caching across app "pauses" may not work.

Thanks @jgainerdewar this is helpful (and a bit surprising).  My previous understanding is pausing Cromwell is just stoping GKE cluster. When starting cluster(PD is not deleted), Cromwell config and it's persistence disk should remains the same including the data that are stored in PD. 
Why "we don't currently persist data in the database"? If so, is there something we can support/improve. Call caching across app "pauses" is an pretty important feature that we want to support

ruchim:
I'm going to ask the question slightly different from Yonghao. If the Cromwell App has a server that is connected to it's own database (lets call it "cromwell-db"), and caches are being written to it because call caching is enabled --> then when the server is stopped and restarted, the server would reconnect to cromwell-db and be able to find cache matches and use them -- does that seem reasonable?

jgainerdewar:
Yes, the question is whether the disk storing the postgres data persists across app restarts - Leo has machinery to enable that, but I'm not sure we're using it.

aednichols:
It's not using a fully-managed database, it's a Docker image spun up alongside the app, and therefore subject to the removal of its persistent storage as Janet describes.

yonghaoy:
Thanks. 
Short summary: 

- Leo apps on GCP creates a detachable persistent disks but CaaA is not using that. 
- Because of this, user will lost historical workflows when pausing/resuming Cromwell. 
- My PR enables call caching within a running instance of the app, and that's is scope of this change. If this PR LG, can I get extra thumb and push this PR forward
- We still want CaaA on GCP use that persistent  if possible. I will hand over this to @ruchim to follow up.

aednichols (approved):
👍 seeing as leaving the app always-on is a supported configuration

But, periodic upgrades for mandatory security updates will cause restart/removed data, so no work with a retention requirement should use it

PR #209, opened 2023-03-15
WX-980 Upgrade TES to 4.0 release

jgainerdewar:


coa-helm/values.yaml
```diff
@@ -13,7 +13,7 @@ fullnameOverride: RUNTIME_PARAMETER
 
 tes: 
   # Test image, should be replaced with an official release when available
```
salonishah11, line 15: nit: Maybe this comment can be removed now?

salonishah11 (approved):

coa-helm/values.yaml
```diff
@@ -12,8 +12,7 @@
 fullnameOverride: RUNTIME_PARAMETER
 
 tes: 
-  # Test image, should be replaced with an official release when available
-  image: mcr.microsoft.com/ga4gh/tes:int-4.0.23005.6784
+  image: mcr.microsoft.com/ga4gh/tes:4.0.23055.6834
```
aednichols, line 15: There are tradeoffs so not clearly right or wrong, but it looks like `mcr.microsoft.com/ga4gh/tes:4` exists and would allow automatic minor version updates.

Right now its SHA256 is identical to `4.0.23055.6834` at `91bd06ea173b27bfc5151ef42d7bbd2b448eaf638e94f733f223437e0da2e230`.

aednichols (approved):

coa-helm/values.yaml
jgainerdewar: Yeah - I think we need to pin to a specific version because we don't have control over their release process and don't want our users to get silent updates we don't know about. We should figure out a way to automatically pull new versions in a compliant way, though.

aednichols: It's not exactly analogous but FWIW we have always used "latest stable" for our PAPI image [(related JIRA)](https://broadworkbench.atlassian.net/browse/CROM-6200).

PR #226, opened 2023-04-12
[WM-1856] Update Readme to add brief info about coa-helm and cromwell-helm chart

salonishah11:
The main purpose of this PR is to make a non-essential change and merge it to Cromwhlem repo so that the Github actions can run successfully and create a Leo PR to update CBAS UI version so that the linked ticket changes can be merged to Dev.

Closes https://broadworkbench.atlassian.net/browse/WM-1856

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,12 @@ while we're iterating on the early helm charts.
 
 ## Charts
 
+### coa-helm:
+Chart for deploying Cromwell as an App on Azure and uses the Composite Batch Analysis Service (C-BAS) as web interface.
+
+### cromwell-helm:
+Chart for deploying Cromwell as an App on GCP and uses the Composite Batch Analysis Service (C-BAS) as web interface.
+
 ### CromwellPostgresMinikube:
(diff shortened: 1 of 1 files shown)


mspector (approved):
these are reasonable changes to the readme on its own, but maybe we should add a `workflow_dispatch` trigger to the action that you want to execute, so that we don't have to submit a PR whenever we need to do this?

salonishah11:
@mspector that's a good point. Since currently the Github actions fail because there is no ticket number assigned to the commit, would adding the `workflow_dispatch` trigger solely work or do we need also to add more logic so that when can provide a ticket number when a manually running the workflow?

mspector:
I just took a look at the action - yeah, I think a small bit of logic would be needed. Maybe it's worth another ticket, if you're in a hurry to get this version into cromwhelm

kpierre13 (approved):

salonishah11:
@mspector yeah I think it would make sense to do it as part of a separate ticket. I will create that ticket.

PR #232, opened 2023-04-14
AJ-939 add source workspaceid

calypsomatic:
As part of [AJ-939](https://broadworkbench.atlassian.net/browse/AJ-939), WDS needs to receive a sourceWorkspaceId. This is passed into the helm chart in https://github.com/DataBiosphere/leonardo/pull/3281.

[AJ-939]: https://broadworkbench.atlassian.net/browse/AJ-939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


coa-helm/values.yaml
```diff
@@ -8,9 +8,11 @@
   batchAccountKey: RUNTIME_PARAMETER
   batchNodesSubnetId: RUNTIME_PARAMETER
   drsUrl: RUNTIME_PARAMETER
+  sourceWorkspaceId: RUNTIME_PARAMETER
```
calypsomatic, line 11: I put this under `config` pretty much randomly; is there a better place for it?  Does it matter?

davidangb: I don't feel strongly about this either way; I think config is fine. I might suggest an entirely new section "provenance" to handle any future values of "where does this data come from?" but it might end up being a new section with only one value.

coa-helm/values.yaml
davidangb, line 1: I think this needs the same updates in [terra-batch-libchart/values.yaml](https://github.com/broadinstitute/cromwhelm/blob/4768e3c31cd247c1a6dfbab49b9c5294ab67b0e4/terra-batch-libchart/values.yaml#L43), too

coa-helm/values.yaml
yuliadub: would it make sense to make a clone section in case in the future there are more values that we need to pass (like the short term token?)

calypsomatic: would clone fall under provenance?  so that if i create a `provenance` section everything for cloning can go under it as well?

coa-helm/values.yaml
calypsomatic: I can put it there just to be safe, but the reason why I didn't was because I saw that `applicationInsightsConnectionString` is in `coa-helm/values.yaml` but not in `terra-batch-libchart/values.yaml` and seems to make it to wds just fine.

davidangb: huh. If it's working empirically, that's fine with me - no need to change. I get lost in these charts :(

yuliadub (approved):

terra-batch-libchart/values.yaml
```diff
@@ -43,6 +43,9 @@ exports:
     config:
       resourceGroup: "UNDEFINED"
 
+    provenance:
+      sourceWorkspaceId: ""
```
calypsomatic, line 47: This needs to be blank or else when not a clone the value will be injected as the default "RUNTIME_PARAMETER"

davidangb: should it change to "" in coa-helm/values.yaml too?

davidangb (approved):
one comment, otherwise :+1

PR #234, opened 2023-04-18
Adding userAccessToken for WDS/COA

lucymcnatt:


terra-batch-libchart/templates/_wds.tpl
```diff
@@ -59,6 +59,8 @@ spec:
               value: {{ include "app.fullname" . }}
             - name: CLIENT_ID
               value: "{{ .Values.identity.clientId }}"
+            - name: USER_ACCESS_TOKEN
+              value: "{{ .Values.provenance.userAccessToken }}"
```
LizBaldo, line 63: Lurking around previous PRs, you might need to update the terra-batch-libchart values.yaml as well https://github.com/broadinstitute/cromwhelm/pull/232/files

lucymcnatt: yeah I was wondering about that myself because I saw the comment 

> any values shared across the GCP and Azure app types should be in terra-batch-libchart, it is right and good for Azure-specific concerns like applicationInsightsConnectionString to live only in this chart

AFAIK this is only Azure at the moment so I shouldn't put it in terra-batch-libchart, or am I incorrect there?

calypsomatic: That was my thought too, but it failed the helm check when I left it out of terra-batch-libchart, so 🤷 I think it should be there (sorry I forgot when I approved!)

PR #237, opened 2023-04-20
AJ-916 Clean up shared CBAS-WDS helm logic for standalone WDS

yuliadub:
Clean up to enable standalone WDS, ticket link [here](https://broadworkbench.atlassian.net/browse/AJ-916).

wds chart now will live here: https://github.com/broadinstitute/terra-helmfile/tree/master/charts/wds


coa-helm/templates/cbas-ui.yaml
```diff
@@ -27,8 +27,6 @@ data:
     {
       "leoUrlRoot" : "{{ .Values.leonardo.url }}",
       "cbasUrlRoot" : "{{ .Values.relay.path }}/cbas",
-      "cromwellUrlRoot" : "{{ .Values.relay.path }}/cromwell",
-      "wdsUrlRoot" : "{{ .Values.relay.path }}/wds",
-      "wdsInstanceId" : "{{ .Values.persistence.workspaceManager.workspaceId }}"
+      "cromwellUrlRoot" : "{{ .Values.relay.path }}/cromwell"
```
calypsomatic, line 30: I think cbas will still need these wds values in order to communicate with it.

cromwell-helm/templates/cbas-ui.yaml
```diff
@@ -22,8 +22,6 @@ data:
   {{ .Values.cbasUI.conf_file }}: |-
     {
       "cbasUrlRoot" : "{{ .Values.ingress.path }}/cbas",
-      "cromwellUrlRoot" : "{{ .Values.ingress.path }}/cromwell",
-      "wdsUrlRoot" : "{{ .Values.ingress.path }}/wds",
-      "wdsInstanceId" : "{{ .Values.persistence.workspaceManager.workspaceId }}"
+      "cromwellUrlRoot" : "{{ .Values.ingress.path }}/cromwell"
```
calypsomatic, line 25: Ditto here

aaronkanzer:
Other than @calypsomatic comments -- this LGTM

coa-helm/templates/cbas-ui.yaml
yuliadub: I wasnt sure if the .Values.relay.path is still the right url root? after digging around I guess the url hasnt changes, just something else is deploying it, but in this context I guess I am not quite sure if it is safe to assume that Values.relay and Values.ingess are going to have the right path for wds now that the charts are decoupled?

davidangb (approved):
I think this looks right, definitely want to hear from the Workflow(s) team(s) on this one

cjllanwarne (approved):

PR #249, opened 2023-05-01
AJ-1005: Pin postgres to 15.2

aaronkanzer:
https://broadworkbench.atlassian.net/browse/AJ-1005

Just for reliability -- pin Postgres to current working version as of the day of this PR for Terra on Azure

### DIFF
--- a/terra-batch-libchart/values.yaml
+++ b/terra-batch-libchart/values.yaml
@@ -63,7 +63,7 @@ exports:
       url: "UNDEFINED"
 
     postgres:
-      image: postgres
+      image: postgres:15.2
       port: 5432
       user: postgres_user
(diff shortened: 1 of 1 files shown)


jgainerdewar (approved):
LGTM but will need a Jira ticket number for compliance.

rtitle (approved):
👍  Perhaps make the same change for WDS? https://github.com/broadinstitute/terra-helmfile/blob/master/charts/wds/values.yaml#L29

aaronkanzer:
> 👍 Perhaps make the same change for WDS? https://github.com/broadinstitute/terra-helmfile/blob/master/charts/wds/values.yaml#L29

Ah good point -- will update in separate PR

PR #264, opened 2023-06-13
AJ-1084: Set leoServiceName label on Cromwell, TES, CBAS, CBAS-UI pods for metrics

rtitle:
Relates to https://github.com/DataBiosphere/leonardo/pull/3423


coa-helm/templates/tes.yaml
```diff
@@ -17,6 +17,8 @@ spec:
       labels:
         io.kompose.service: tes
         aadpodidbinding: {{ .Values.identity.name }}
+        leoServiceName: tes
+    {{ include "coa-helm.labels" $ | nindent 8 }}
```
rtitle, line 21: TES didn't have the standard labels including `leoAppName`, `workspaceId` -- so added that as well.

cjllanwarne: This indent looks different than in the other files but `nindent` is still 8. Is that right?

cjllanwarne (approved):
LGTM assuming the indent is right

jgainerdewar (approved):

PR #265, opened 2023-07-17
WX-1199 LZ postgres and workload identity

jgainerdewar:
This PR does a lot of setup to enable LZ postgres for the Workflow app, but IS NOT sufficient to actually use it. In particular, we don't yet have a TES build with AAD Postgres auth. I tested:
 * **Deploy with Leo config unchanged.** App functions normally.
 * **Deploy with Leo config to enable LZ postgres.** Cromwell and CBAS are happy and can do database operations, TES can't boot because it has no database.
 * **Deploy hacked-up chart that lets TES have docker Postgres, puts Cromwell and CBAS in LZ postgres** App functions normally.

Also adds configmap hashes for Cromwell, CBAS, CBAS UI, NGINX so that pods will automatically restart when config is updated.


dvoet (approved):

coa-helm/templates/cbas-api.yaml
```diff
@@ -28,22 +32,34 @@ data:
   {{ .Values.cbas.conf_file }}: |-
     spring:
       datasource:
+        {{- if .Values.postgres.podLocalDatabaseEnabled }}
         jdbc-url: jdbc:postgresql://{{ include "app.fullname" . }}-postgres:{{ .Values.postgres.port }}/{{ .Values.postgres.cbas.dbname }}
-        username: {{ .Values.postgres.user }}
         password: {{ include "dbPassword" . | b64enc | quote }}
+        {{- else }}
+        jdbc-url: "jdbc:postgresql://{{ .Values.postgres.host }}:{{ .Values.postgres.port }}/{{ .Values.postgres.cbas.dbname }}?sslMode=Require&authenticationPluginClassName=com.azure.identity.extensions.jdbc.postgresql.AzurePostgresqlAuthenticationPlugin"
```
cjllanwarne, line 39: My DRY senses are tingling, especially since this is 99% coming from `values` anyway, but I'm not sure I see a neat way to do it without a new utility function, and I wouldn't suggest doing that for this PR

cjllanwarne (approved):

cjllanwarne:
There is also a line in `persistence.yaml` that you can make dependent on postgres being enabled or not

coa-helm/values.yaml
```diff
@@ -70,3 +70,10 @@ identity:
   name: RUNTIME_PARAMETER
   resourceId: RUNTIME_PARAMETER
   clientId: RUNTIME_PARAMETER
+
+workloadIdentity:
+  enabled: false
+  serviceAccountName: RUNTIME_PARAMETER
+
+postgres:
+  podLocalDatabaseEnabled: true
```
jgainerdewar, line 79: This is a repetition of the value in the library chart but I think it's convenient to include here for readability and potentially setting a new default for this chart.

coa-helm/templates/tes.yaml
```diff
@@ -105,20 +107,33 @@ spec:
...
+            {{- else -}}
+            - name: TesPostgreSql__ServerName
+              value: {{ .Values.postgres.host | quote }}
+            - name: TesPostgreSql__ServerNameSuffix
+              value: ""
+            - name: TesPostgreSql__DatabaseUserPassword
+              value: ""
+            - name: TesPostgreSql__UseManagedIdentity
+              value: "true"
+            {{- end }}
```
jgainerdewar, line 136: Note that the `podLocalDatabaseEnabled=false` path is UNTESTED and not functional with the current TES version.

PR #270, opened 2023-08-24
WX-1245 WX-1258 PgBouncer Support, allow TES only in local postgres

jgainerdewar:
This PR makes two separate changes to bring `coa-helm` into its pre-audit final form:
 * Allow Cromwell and CBAS to move to LZ Postgres while TES stays in local docker postgres. 
 * Choose postgres port based on whether we're in LZ Postgres with PgBouncer enabled or not.


THWiseman (commented):
LGTM

coa-helm/values.yaml
```diff
@@ -77,3 +77,7 @@ workloadIdentity:
 
 postgres:
   podLocalDatabaseEnabled: true
+  podLocalDatabaseEnabledForTES: true
+  pgbouncer:
+    enabled: false
```
aednichols, line 82: Should this be `true` as per the comment "can be set to true in coa-helm" or is that a future thing?

jgainerdewar: It should be false by default here, and will be set to true by Leo in cases where PgBouncer exists.

THWiseman (approved):

PR #272, opened 2023-08-28
WX-1261 Only use LZ Postgres in Azure if PgBouncer is enabled

jgainerdewar:


coa-helm/templates/cbas-api.yaml
```diff
@@ -34,7 +34,7 @@ data:
   {{ .Values.cbas.conf_file }}: |-
     spring:
       datasource:
-        {{- if .Values.postgres.podLocalDatabaseEnabled }}
+        {{- if eq ("true") (include "podLocalDatabaseEnabledForCromwellCBAS" .) }}
```
dillydally414, line 37: Do we need an `if eq (true) (enabled)` or can we just do `if (enabled)`?

jgainerdewar: You would think so! It seems that Helm helper functions like these cannot directly output booleans, very annoying. 

https://github.com/helm/helm/issues/10382

terra-batch-libchart/templates/_helpers.tpl
```diff
@@ -92,14 +92,44 @@ Lookup the existing secret values if they exist, or generate a random value
...
 {{- define "postgresPort" -}}
-{{- if and (.Values.postgres.pgbouncer.enabled) (not (.Values.postgres.podLocalDatabaseEnabled)) }}
+{{- if (eq ("true") (include "useLZPostgresWithPgBouncer" .)) }}
 {{- .Values.postgres.pgbouncer.port }}
 {{- else }}
 {{- .Values.postgres.port }}
 {{- end }}
 {{- end -}}
+
+{{/*
```
dillydally414, line 115: Can you explain the difference between the 3 flags added to this file? I see they're used in different places, but it seems like they are all connected to the same values.

jgainerdewar: This was my effort to make things clearer in all the parts of the chart that use these values. As you point out, they're all connected to the same values, but they're using them to answer different questions. 

On Azure, there are currently three valid database layouts to run this app with:
 * Everything in pod-local Postgres (this is what we have been doing up until now)
 * CBAS and Cromwell in LZ Postgres, TES in pod-local Postgres (this is what we're going to turn on soon)
 * Everything in LZ Postgres (this one is supported by the chart but not TES, so we aren't actually using it yet)
Depending on which of those we're doing, we need to decide whether to deploy Postgres as part of the app or not, and decide how to tell each component to connect to its database. It tuns out to be pretty gross to decide each of those based on the chart values, so I added these flags to make those decisions easier to understand.

 * `useLZPostgresWithPgBouncer` is just used by other functions in this file. It defines the conditions under which we want to run with LZ postgres with PgBouncer at all.
 * `podLocalDatabaseEnabledForCromwellCBAS` is a way to express the above in language that makes more sense with the rest of chart
 * `buildPodLocalDatabase` tells us whether to create the pod-local database in the app.

dillydally414: Gotcha, thanks for clarifying!

dillydally414 (approved):

rtitle (approved):
Logic looks good to me -- ctmpl boolean logic is something else though 😅

coa-helm/templates/cbas-api.yaml
rtitle: Source: https://github.com/hashicorp/consul-template/blob/main/docs/templating-language.md

PR #286, opened 2024-04-03
[WM-2409] Remove unused templates in cromwell-helm chart for GCP

salonishah11:
This PR removes from GCP cromwell charts :
- no longer needed cbas-api.yaml file
- job-manager.yaml - JM doesn't work in Cromwell app in GCP

Closes https://broadworkbench.atlassian.net/browse/WM-2409


cromwell-helm/values.yaml
```diff
@@ -43,75 +43,5 @@ cromwell:
     dir: /etc/conf
     file: papi.conf
 
-jobmanager:
```
salonishah11, line 46: Question for reviewers: This can now be removed right? I wasn't able to find cromwell-helm references for this and hence removed it but I am not sure if it is used else where.

jgainerdewar: Yeah, if it's not used here seems fine to remove. Might want to double-check in Leo that it isn't trying to override any of it? I doubt it but doesn't hurt to be safe.

kpierre13 (approved):

cromwell-helm/values.yaml
salonishah11: I wasn't able to find any job manager reference in [CromwellAppInstall](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/app/CromwellAppInstall.scala).

jgainerdewar (approved):

PR #288, opened 2024-06-10
RW-12676 add resource request/limit

Qi77Qi:
1. Add an `autopilot.enabled` flag
2. Add `resources` section if `autopilot.enabled` is true.


jgainerdewar (approved):

cjllanwarne (approved):
Not really something I maintain any more but looks good at first glance. I assume this is not going to hurt anything if deployed to the _other_ cloud?

PR #289, opened 2025-04-15
[AN-439] Add GCPBATCH backend config and have the backend default configurable

LizBaldo:
JIRA Ticket: https://broadworkbench.atlassian.net/browse/AN-439


cromwell-helm/templates/cromwell.yaml
```diff
@@ -166,6 +166,54 @@ data:
...
+                auth = "application-default"
+                // Google project which will be billed for the requests
+                project = "{{ .Values.config.gcsProject }}"
+              }
+            }
+            virtual-private-cloud {
+              network-label-key = "vpc-network-name"
+              subnetwork-label-key = "vpc-subnetwork-name"
+              auth = "application-default"
+            }
```
aednichols, line 214: Batch in Terra is [a bit different here](https://github.com/broadinstitute/terra-helmfile/blob/88dc8a48684dafebf6ba2f213a27d398332badee/charts/cromwell/templates/config/_cromwell-gcp-batch.conf.tpl#L43-L47), not sure if it applies to this use case as well.

aednichols (approved):

cromwell-helm/templates/cromwell.yaml
LizBaldo: Good catch, thanks! I was basing this on the `GCPBATCH.conf` file in the Cromwell repo, but the actual helm file template config is likely more accurate than the example configs

aednichols: Ah yeah, unfortunately that may be out of date or intended for a generic not-Terra installation.

cromwell-helm/values.yaml
```diff
@@ -1,8 +1,8 @@
 config:
   gcsBucket: "SUPPLIED BY LEO"
   gcsProject: "SUPPLIED BY LEO"
-  gcsRegion: us-central1 # TODO: SUPPLY VIA LEO (BW-924)
-  papiv2EndpointUrl: https://lifesciences.googleapis.com/ # TODO: SUPPLY VIA LEO... or hard-code into Cromwell? (BW-924)
+  gcsRegion: "SUPPLIED BY LEO"
```
jgainerdewar, line 4: Is Leo already supplying this and we're just not using it? Wondering if it's easier to leave this as-is.

LizBaldo: Oh we are using it, I already modified leo to supply it, testing now.

aednichols (approved):

lucymcnatt (approved):

cromwell-helm/templates/cromwell.yaml
```diff
@@ -123,18 +123,12 @@ data:
...
-              // token = ""
-            }
-
             genomics {
               // A reference to an auth defined in the `google` stanza at the top.  This auth is used to create
               // Pipelines and manipulate auth JSONs.
               auth = "application-default"
               // Endpoint for API (dependent on 'location')
-              endpoint-url = "{{ .Values.config.papiv2EndpointUrl }}"
+              endpoint-url = "https://lifesciences.googleapis.com/"
```
lucymcnatt, line 131: is the endpoint url different for papi vs. batch? ie. should this be dependent on the value from Leo?

cromwell-helm/templates/cromwell.yaml
```diff
@@ -166,6 +160,48 @@ data:
...
+
+            filesystems {
+              gcs {
+                // A reference to a potentially different auth for manipulating files via engine functions.
+                auth = "application-default"
+                // Google project which will be billed for the requests
+                project = "{{ .Values.config.gcsProject }}"
+              }
+            }
+            // Cromwell will replace * with the correct region for each job submission.
```
LizBaldo, line 198: @aednichols and/or @salonishah11  happy to chat tomorrow, but this does not seem to be the case. When trying to test this I am getting this very straightforward error:

``` Unable to complete Batch request due to a problem with the request (io.grpc.StatusRuntimeException: INVALID_ARGUMENT: subnetwork field is invalid. subnetwork: projects/terra-quality-e438a884/regions/*/subnetworks/subnetwork is not matching the expected format: regions/([a-z0-9-_]+)/subnetworks/([a-z]([-a-z0-9]*[a-z0-9])?)$). 
```

LizBaldo: In this case, we likely want to replace the `*` with a {{ .Values.config.gcsRegion }}

LizBaldo: Actually, this might instead mean that the cromwell chart needs to point to the latest Cromwell release 🤔

cromwell-helm/templates/cromwell.yaml
LizBaldo: Batch does not seem to need this endpoint, it is only required in the papi config

cromwell-helm/templates/cromwell.yaml
LizBaldo: Yep, I think I need to update this: https://github.com/broadinstitute/cromwhelm/blob/0f5e53b8488e0fa2f2614826920635af780f57dc/terra-batch-libchart/values.yaml#L5

cromwell-helm/templates/cromwell.yaml
lucymcnatt: ah gotcha, thanks for the clarification!

lucymcnatt (approved):

cromwell-helm/templates/cromwell.yaml
LizBaldo: Things are now working when pointing to cromwell:89-561bbf4

PR #293, opened 2025-10-09
[RW-16198] remove papi reference

Qi77Qi:
https://precisionmedicineinitiative.atlassian.net/browse/RW-16198


aednichols (approved):
LGTM 👍 thank you!

LizBaldo (approved):
Good catch, thank you!

